PR #3, opened 2018-11-29
[llvm] Add IndexStoreDB_ prefix to LLVMSupport to avoid conflicts

benlangmuir:
SwiftPM doesn't protect us from module name conflicts, and this happens
to collide with llbuild's names. Add a prefix to workaround.

See https://bugs.swift.org/browse/SR-9292


akyrtzi:
Reminder to update the pin file.

PR #4, opened 2018-12-19
[config] Disable futimens to fix backward deployment issue

benlangmuir:
This API is available in macOS 10.13, but since this configuration is
fixed, we should be conservative. In practice the only effect this has
is to fix a warning while building this code on macOS, because the only
function that uses futimens is setLastModificationAndAccessTime, which
is not used anywhere in indexstore-db.


benlangmuir:
I'll update sourcekit-lsp as well once this is merged to avoid the warning.

PR #8, opened 2019-02-07
[IndexDatastore] For the out-of-date trigger notifications report all the out-of-date files without any coalescing

akyrtzi:
Previously if there were multiple headers that got out-of-date and they were all included by the same main file, it would trigger a single notification for that main file. With these changes there will be notifications for each file.

This allows clients to get the full information of what needs to get updated and they get the responsibility for doing any kind of coalescing they want.

### DIFF
--- a/include/IndexStoreDB/Index/IndexSystemDelegate.h
+++ b/include/IndexStoreDB/Index/IndexSystemDelegate.h
@@ -27,6 +27,7 @@ namespace index {
 class OutOfDateTriggerHint {
 public:
   virtual ~OutOfDateTriggerHint() {}
+  virtual std::string originalFileTrigger() = 0;
   virtual std::string description() = 0;
 
 private:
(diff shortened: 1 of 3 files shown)


benlangmuir:
Can you say more about how this should work end-to-end? From a quick read, it seems like we have already coalesced multiple changes to a unit, then we split them up into multiple notifications, and then the client would probably want to coalesce them back together again.  Is there an advantage to having multiple notifications like this, instead of changing the notification to contain multiple changes?

akyrtzi:
> instead of changing the notification to contain multiple changes?

That would be more complicated. Right now the changed notifications are reported immediately as the datastore becomes aware of them. Having a single notification with multiple changes for a unit would involve more complicated coalescing, potentially involving timers, and coalescing at the unit level only would not be enough, otherwise the client would still need to do global coalescing anyway.

I find it a better model to keep indexstore a simple notification producer and leave the complicated coalescing logic at the client level.

PR #12, opened 2019-03-21
Remove all Uses of TimeValue

jmittert:
The latest version of llvm no longer has TimeValue.h, so this replaces uses
with TimePoint and std::chrono.

I'd like to start porting this to Windows and as part of that, I need to pull in the Windows headers which means updating the llvm support libraries that are part of indexstore-db.

### DIFF
--- a/lib/Database/ImportTransaction.cpp
+++ b/lib/Database/ImportTransaction.cpp
@@ -158,13 +158,14 @@ IDCode ImportTransaction::Implementation::addModuleName(StringRef moduleName) {
 }
 
 void ImportTransaction::Implementation::addFileAssociationForProvider(IDCode provider, IDCode file, IDCode unit,
-                                                                      llvm::sys::TimeValue modTime, IDCode module, bool isSystem) {
+                                                                      llvm::sys::TimePoint<> modTime, IDCode module, bool isSystem) {
   auto &db = DBase->impl();
   auto &dbiFilesByProvider = db.getDBITimestampedFilesByProvider();
 
   auto cursor = lmdb::cursor::open(Txn, dbiFilesByProvider);
 
-  TimestampedFileForProviderData entry{file, unit, module, modTime.seconds(), modTime.nanoseconds(), isSystem};
+  uint64_t nanos = std::chrono::duration_cast<std::chrono::nanoseconds>(modTime.time_since_epoch()).count();
+  TimestampedFileForProviderData entry{file, unit, module, nanos, isSystem};
   lmdb::val key{&provider, sizeof(provider)};
   lmdb::val value{&entry, sizeof(entry)};
   bool added = cursor.put(key, value, MDB_NODUPDATA);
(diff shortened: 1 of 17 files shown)


lib/Database/ImportTransaction.cpp
```diff
@@ -158,13 +158,16 @@ IDCode ImportTransaction::Implementation::addModuleName(StringRef moduleName) {
...
+                                                                      llvm::sys::TimePoint<> modTime, IDCode module, bool isSystem) {
   auto &db = DBase->impl();
   auto &dbiFilesByProvider = db.getDBITimestampedFilesByProvider();
 
   auto cursor = lmdb::cursor::open(Txn, dbiFilesByProvider);
 
-  TimestampedFileForProviderData entry{file, unit, module, modTime.seconds(), modTime.nanoseconds(), isSystem};
+  auto nanos = std::chrono::duration_cast<std::chrono::nanoseconds>(modTime.time_since_epoch()).count();
+  auto secComponent = nanos / 1000000000;
+  int32_t nanoComponent = nanos % 1000000000;
```
benlangmuir: This changes the interpretation of any existing index data.  TimeValue's internal epoch was/is `00:00:00 Jan 1, 2000 UTC`, but this changes it to store the time using `system_clock`'s epoch, which is probably the posix epoch in 1970.

I think we need to bump `DATABASE_FORMAT_VERSION` CC @akyrtzi to verify this.

@akyrtzi since we need to bump the format anyway, is there a reason we need to split up the time into components like this?  Could we store a `std::chrono::time_point<std::chrono::system_clock, nanoseconds>` directly instead (we'd want to use that instead of `TimePoint<>` to guard against changes in llvm)?

lib/Database/ImportTransaction.cpp
```diff
@@ -174,7 +177,8 @@ void ImportTransaction::Implementation::addFileAssociationForProvider(IDCode pro
     lmdb::val existingValue;
     cursor.get(existingKey, existingValue, MDB_GET_CURRENT);
     const auto &existingData = *(TimestampedFileForProviderData*)existingValue.data();
-    llvm::sys::TimeValue existingModTime(existingData.Seconds, existingData.Nanoseconds);
+    llvm::sys::TimePoint<> existingModTime = std::chrono::time_point_cast<std::chrono::nanoseconds>(llvm::sys::toTimePoint(existingData.Seconds));
```
benlangmuir: `toTimePoint` assumes you're converting from a `time_t`.  We should only do that if we're using `toTimeT` on the other side.

lib/Database/ImportTransaction.cpp
akyrtzi: > @akyrtzi since we need to bump the format anyway, is there a reason we need to split up the time into components like this? Could we store a `std::chrono::time_point<std::chrono::system_clock, nanoseconds>` directly instead

I agree, we should store this as a single uint64 and bump the database format.

jmittert:
Updated the database version and now use `uint64_t NanoTime` instead of Seconds and Nanoseconds. Assuming I did the math right, nanoseconds since the epoch fits in a uint64_t for the next few hundred years.

benlangmuir:
@swift-ci please test

PR #13, opened 2019-03-21
Remove LLVM_EXPORT references

jmittert:
LLVM_EXPORT no longer exists in llvm, remove it in preparation to pull in a new version

### DIFF
--- a/include/IndexStoreDB/Support/Visibility.h
+++ b/include/IndexStoreDB/Support/Visibility.h
@@ -1,0 +1,26 @@
+//===--- Visibility.h ------------------------------------------*- C++ -*-===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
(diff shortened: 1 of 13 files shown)


jmittert:
> We want to be able to use this code with `-fvisibility=hidden`, so we should add a new macro to replace this. I suggest the name `INDEXSTOREDB_EXPORT`, which should be okay since it's not used in very many places.

Fair enough!

include/IndexStoreDB/Support/Visibility.h
```diff
@@ -0,0 +1,22 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_INDEXSTOREDB_SUPPORT_VISIBILITY_H
+#define LLVM_INDEXSTOREDB_SUPPORT_VISIBILITY_H
+
+#if __has_attribute(visibility)
```
benlangmuir:
```
#ifndef __has_attribute
#define __has_attribute(x) 0
#endif
```

jmittert: Ah right, added.

benlangmuir:
@swift-ci please test

benlangmuir:
Build is running, but the status isn't showing up for some reason.  I'll verify it manually once it's done.

PR #14, opened 2019-03-21
Update the Import Script to Include Windows Headers

gmittert:
Update the import script and run it to import a new version of llvm. The only file changes are in the first commit. Everything in the second commit should be identical to swift llvm commit ab94816b593daf1c623322b0fb981abb0b156833 except for llvm/Support/Process.cpp with required some hacks in the import-llvm script.


jmittert:
Indexstore includes `llvm/ADT/OptionSet.h`, which from as far as I can tell is from [this patch](https://reviews.llvm.org/D17325) that never got merged. I've just copied it out and have the import script copy it back in. However really, either the uses of it should be removed, or it should be necroed an merged it.

benlangmuir:
Instead of moving the import script, I suggest using a different name for the directory containing the files.  That will also make it a lot easier to review the changes to that script.

jmittert:
Alright, modified this one to just be the import script changes. I should be much easier to read now.

Utilities/import-llvm
```diff
@@ -15,41 +15,55 @@ import sys
...
-ADT_imports += ['DenseMap', 'DenseMapInfo', 'DenseSet', 'FoldingSet', 'StringMap', 'StringSet']
+ADT_imports += ['DenseSet', 'DenseMap', 'DenseMapInfo', 'FoldingSet', 'StringMap', 'StringSet']
 
 # ADT "Small" structures.
 ADT_imports += ['SmallPtrSet', 'SmallSet', 'SmallString', 'SmallVector']
 
 # ADT Algorithms.
 ADT_imports += ['edit_distance']
 
+Demangle_imports = ['Demangle', 'ItaniumDemangle', 'MicrosoftDemangle',
```
benlangmuir: What pulled in the demangler?

Utilities/import-llvm
```diff
@@ -190,6 +235,51 @@ def main():
...
+    os.symlink(os.path.join('..', '..', '..', '..', 'include', 'llvm-c'),
+               os.path.join(sourcekit_llvm_lib, 'Support', 'include', 'llvm-c'))
+    os.symlink(os.path.join('..', '..', '..', '..', 'include', 'llvm'),
+               os.path.join(sourcekit_llvm_lib, 'Demangle', 'include', 'llvm'))
+    os.symlink(os.path.join('..', '..', '..', 'llvm'),
+               os.path.join(sourcekit_llvm_include, 'Demangle', 'include', 'llvm'))
+    os.symlink(os.path.join('..', '..', '..', 'include', 'llvm'),
+               os.path.join(sourcekit_llvm_lib, 'include', 'llvm'))
+    os.symlink(os.path.join('..', '..', '..', 'include', 'llvm-c'),
+               os.path.join(sourcekit_llvm_lib, 'include', 'llvm-c'))
```
benlangmuir: Could these symlinks be done by looping over the directories instead of listing them out manually?  It's really hard to tell if these are correct by inspection.

Utilities/import-llvm
```diff
@@ -190,6 +235,51 @@ def main():
...
+    # AArch64TargetParser uses a relative path to find the .def file, so we
+    # copy this one to the expected location :(
+    def_location = os.path.join(sourcekit_srcroot, 'lib', 'include', 'llvm', 'Support')
+    if os.path.exists(def_location):
+        shutil.rmtree(def_location)
+    os.makedirs(def_location)
+    copyfile(os.path.join(sourcekit_srcroot, 'include', 'llvm', 'Support', 'AArch64TargetParser.def'),
+             os.path.join(def_location, 'AArch64TargetParser.def'))
+
+    # Process.cpp uses LLVM_ENABLE_CRASH_DUMPS which is usually set in CMake.
```
benlangmuir: This should be in config.h.  In general, it would be good to look at a new config.h from an llvm build, although it will need to be merged manually with what's here, which is already customized.  But that more general change doesn't need to happen in this PR.

Utilities/import-llvm
```diff
@@ -190,6 +235,51 @@ def main():
...
+    copyfile(os.path.join(sourcekit_srcroot, 'Utilities', 'import-llvm.d', 'include', 'llvm', 'ADT', 'OptionSet.h'),
+             os.path.join(sourcekit_srcroot, 'include', 'llvm', 'ADT', 'OptionSet.h'))
+
+    # AArch64TargetParser uses a relative path to find the .def file, so we
+    # copy this one to the expected location :(
+    def_location = os.path.join(sourcekit_srcroot, 'lib', 'include', 'llvm', 'Support')
+    if os.path.exists(def_location):
+        shutil.rmtree(def_location)
+    os.makedirs(def_location)
+    copyfile(os.path.join(sourcekit_srcroot, 'include', 'llvm', 'Support', 'AArch64TargetParser.def'),
```
benlangmuir: Does the original location get looked at? If not, we could move instead of copy.

Utilities/import-llvm
```diff
@@ -15,41 +15,55 @@ import sys
 import subprocess
 
 # ADT types.
-ADT_imports = ['EpochTracker', 'iterator', 'iterator_range', 'Hashing', 'None', 'Optional',
-               'Statistic', 'StringExtras', 'STLExtras', 'AllocatorList', 'Triple']
+ADT_imports = ['iterator', 'iterator_range', 'APFloat', 'APInt', 'APSInt',
+               'AllocatorList', 'EpochTracker', 'Hashing', 'None', 'Optional',
+               'Statistic', 'StringExtras', 'STLExtras']
```
benlangmuir: These lists are getting quite messy.  I think we're better off just making each variable into a single list, and putting every dependency on its own line.  That will make it much easier to see changes in the future.  Splitting it up "thematically" like this made more sense when we had fewer imports and you could still conceivable eyeball changes.

Utilities/import-llvm
jmittert: `Support/Unix/Signals.inc` includes `Demangle.h` it uses it to demangle symbols when printing backtraces.

Utilities/import-llvm
jmittert: Move it is!

Utilities/import-llvm
jmittert: Cleaned these up a bit, I didn't need as many as I thought I did.

benlangmuir:
The TimeValue changes seem to be here too.

Utilities/import-llvm
```diff
@@ -15,50 +11,160 @@ import sys
...
-# Stuff we don't want, but have to pull in.
-Support_imports += [
-    'COFF', 'ConvertUTF', 'ConvertUTFWrapper', 'Debug', 'FileSystem',
-    'FileUtilities', 'Host', 'Locale', 'MachO', 'Path', 'Process', 'Program', 'SMLoc',
-    'SourceMgr', 'Signals', 'StringSaver', 'ToolOutputFile', 'TrailingObjects', 'Unicode', 'UnicodeCharRanges',
-    'MemAlloc', 'Chrono', 'FormatProviders', 'FormatVariadic', 'FormatCommon',
-    'FormatVariadicDetails', 'NativeFormatting', 'DJB', 'ReverseIteration', 'MD5',
-    'SmallVectorMemoryBuffer', 'WithColor', 'Options', 'PrettyStackTrace', 'Watchdog',
-    'TargetParser', 'ARMBuildAttributes', 'ARMTargetParser.def', 'AArch64TargetParser.def', 'X86TargetParser.def', 'LineIterator']
+    'AArch64TargetParser',
```
jmittert: Praise be `M-x sort-lines`

jmittert:
> The TimeValue changes seem to be here too.

rebased off master, gone now.

Utilities/import-llvm
benlangmuir: Ah, this was apparently one of the modifications to the files that we warned about at the top of the import script:

    char* d = nullptr;//itaniumDemangle(dlinfo.dli_sname, nullptr, nullptr, &res);

I don't think we want to pull in the demangler.

Utilities/import-llvm
```diff
@@ -15,50 +11,160 @@ import sys
...
+    'Mutex',
+    'MutexGuard',
+    'NativeFormatting',
+    'Options',
+    'Path',
+    'PointerLikeTypeTraits',
+    'Process',
+    'Program',
+    'Recycler',
+    'Regex',
```
benlangmuir: What is pulling in Regex (and all the reg* files below)?  If possible it would be nice to avoid this.

jmittert: Timer -> YAML (so it can print JSON) -> Regex. It's just one function that indexstoredb doesn't use, so it could be commented out/removed.

jmittert:
Alright, added in some super gross sed calls so that Demangle and Regex don't need to be imported.

Utilities/import-llvm
```diff
@@ -189,11 +305,52 @@ def main():
...
+    def_location = os.path.join(sourcekit_srcroot, 'lib', 'include', 'llvm', 'Support')
+    if os.path.exists(def_location):
+        shutil.rmtree(def_location)
+    os.makedirs(def_location)
+    copyfile(os.path.join(sourcekit_srcroot, 'include', 'llvm', 'Support', 'AArch64TargetParser.def'),
+             os.path.join(def_location, 'AArch64TargetParser.def'))
+
+    # Timer.cpp and Statistics.cpp use YAMLTraits to assert some JSON is valid. We don't want to
+    # pull in the YAML/Regex so we'll remove the "offending" lines.
+    timer_location = os.path.join(sourcekit_srcroot, 'lib', 'LLVMSupport', 'Support', 'Timer.cpp')
```
benlangmuir: Yikes, this is pretty gross.  Could we use a patch file instead of doing this with `sed`?

jmittert: Sure, that seems like a much nice way to do it.

jmittert:
Ping on this one

akyrtzi:
Apologies, this PR slipped my mind, and I've already gone ahead and imported a new llvm here: https://github.com/apple/indexstore-db/pull/18

I introduced a [patch file](https://github.com/apple/indexstore-db/blob/master/Utilities/llvm-patch.diff) to avoid adding the regex dependency and to fix include errors (ideally the patch file changes should go to llvm upstream).

I avoided adding the demangle dependency by adding a [dummy Demangle.h](https://github.com/apple/indexstore-db/blob/master/include/llvm/Demangle/Demangle.h)

I also added the Windows support files.
If there is something missing for Windows support please open a new PR against master.

PR #16, opened 2019-03-28
Windows Port

gmittert:
This allows IndexStoreDB to build on Windows.

Once the other PRs go in, this enables compilation on Windows. I have some CMake locally which I've been using to build this, but it should work just as well with SwiftPM when that is ready on Windows.

### DIFF
--- a/include/IndexStoreDB/Database/Database.h
+++ b/include/IndexStoreDB/Database/Database.h
@@ -17,6 +17,7 @@
 #include "IndexStoreDB/Support/LLVM.h"
 #include "IndexStoreDB/Support/Visibility.h"
 #include <memory>
+#include <string>
 
 namespace IndexStoreDB {
 namespace db {
--- a/lib/Database/Database.cpp
+++ b/lib/Database/Database.cpp
@@ -23,6 +23,15 @@
 #include "llvm/Support/Mutex.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/raw_ostream.h"
+#if defined(_WIN32)
+#include "Windows.h"
+#endif
+
+#if defined(_WIN32)
+typedef HANDLE indexstorePid_t;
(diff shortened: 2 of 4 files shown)


gmittert:
@swift-ci please test

gmittert:
I've removed the CrossPlatform.[h/cpp] and instead inlined the changes.

@swift-ci please test

gmittert:
> I left some comments on the canonical path implementation, but actually: can we just use llvm's `sys::fs::real_path`?

Ugh, that was really leaky. I've changed it to use llvm real_path. Like the Unix realpath, if PathBuf is NULL, it'll allocate memory and the target of the stringref would have to be freed. Thoughts on just having it return a std::string? I believe it's only ever used on Path.cpp:86.

benlangmuir:
> Thoughts on just having it return a std::string? I believe it's only ever used on Path.cpp:86.

Ah, I meant why not replace this whole function with a call to `llvm::sys::fs::real_path`?  Instead of passing in a PathBuf char array we can use a `SmallString<PATH_MAX>`.  It will have an extra string copy compared to the current implementation on macOS/Linux, but won't have to heap allocate for any platform.

gmittert:
> 
> Ah, I meant why not replace this whole function with a call to `llvm::sys::fs::real_path`? Instead of passing in a PathBuf char array we can use a `SmallString<PATH_MAX>`. It will have an extra string copy compared to the current implementation on macOS/Linux, but won't have to heap allocate for any platform.

Oh I see, yes, that makes much more sense in that context.

@swift-ci please test

benlangmuir:
Not sure where the PR test status went, but it's still running apparently https://ci.swift.org/job/swift-indexstore-db-PR-macOS/24/  https://ci.swift.org/job/swift-indexstore-db-PR-Linux/25/

PR #17, opened 2019-04-15
[Index] Use a single global serial queue for registration of units

akyrtzi:
This is useful to avoid doing a lot of parallel CPU and I/O work when opening multiple workspaces.
It also allows the unit processing to stop when the workspace closes, instead of continuing until all units have been processed.

rdar://36996503

### DIFF
--- a/lib/Index/IndexDatastore.cpp
+++ b/lib/Index/IndexDatastore.cpp
@@ -61,6 +61,18 @@ static sys::TimePoint<> toTimePoint(timespec ts) {
 
 const static dispatch_qos_class_t unitChangesQOS = QOS_CLASS_UTILITY;
 
+/// Returns a global serial queue for unit processing.
+/// This is useful to avoid doing a lot of parallel CPU and I/O work when opening multiple workspaces.
+static dispatch_queue_t getGlobalQueueForUnitChanges() {
+  static dispatch_queue_t queueForUnitChanges;
+  static dispatch_once_t onceToken = 0;
+  dispatch_once(&onceToken, ^{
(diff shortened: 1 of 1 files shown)


akyrtzi:
@swift-ci Please test

lib/Index/IndexDatastore.cpp
```diff
@@ -707,28 +712,29 @@ bool IndexDatastoreImpl::init(IndexStoreRef idxStore,
...
+      auto cevt = EventNote.getEvent(i);
+      UnitEventInfo evt = UnitEventInfo{cevt.getKind(), cevt.getUnitName()};
+
+      // Create the block with QoS explicitly to ensure that the QoS from the indexstore callback can't affect the onFilesChange priority. This call may do a lot of I/O and we don't want to wedge the system by running at elevated priority.
+      dispatch_block_t onUnitChangeBlock = dispatch_block_create_with_qos_class(DISPATCH_BLOCK_INHERIT_QOS_CLASS, unitChangesQOS, 0, ^{
+        auto UnitRepo = WeakUnitRepo.lock();
+        if (!UnitRepo)
+          return;
+
+        Delegate->processingAddedPending(1);
```
benlangmuir: Putting each event in its own block is unlikely to intermix the events between workspaces, because it still queues up all of the blocks at the beginning using `dispatch_async`.  Unless the workspaces are racing to add events, all the events from workspace 1 get queued before workspace 2 even starts.

If we want to intermix the events, we probably need to be more proactive.  Each workspace could have a list of outstanding events, and the onUnitChange block can pull 1 (or any batch size) events from there at a time and reschedule itself with dispatch_async.  That way other workspaces can add their own block to the queue in between handing events.

akyrtzi: Good point, see changes.

PR #19, opened 2019-04-25
[Index] For registering units process them in batches of maximum 10 at a time

akyrtzi:
This introduces a const variable to control the max number of batches, instead of only being able to process one at a time.


akyrtzi:
@swift-ci Please test

PR #20, opened 2019-04-28
[IndexDatastore] Ensure `IndexDatastore::waitUntilDoneInitializing()` works properly

akyrtzi:
It wasn't taking into account processing the initial units as multiple batches and the case where the data store directory has no units.


akyrtzi:
@swift-ci Please test

PR #21, opened 2019-04-30
[Index] Introduce IndexSystem::foreachUnitTestSymbolReferencedByOutputPaths()

akyrtzi:
This is useful for retrieving all the unit test class/method occurrences that are referenced from a set of output file paths and their associated units.


akyrtzi:
@swift-ci Please test

PR #23, opened 2019-05-01
[Database] For ReadTransaction::getProviderFileCodeReferences() make sure to filter the units properly

akyrtzi:
Previously it was returning the unit with the most recent modification time, and if that unit was filtered-out by SymbolIndexImpl::createProviderForCode()
then the whole provider would be ignored, which is incorrect because there could be other units for the same provider that are eligible.

This change moves the filtering into the `ReadTransaction`'s call.


akyrtzi:
@swift-ci Please test

PR #29, opened 2019-06-29
Add ability to find symbols by name

literalpie:
This can be used in lsp projects to find
all symbols in a workspace that match a string.

https://bugs.swift.org/browse/SR-10806

### DIFF
--- a/Sources/IndexStoreDB/IndexStoreDB.swift
+++ b/Sources/IndexStoreDB/IndexStoreDB.swift
@@ -80,6 +80,33 @@ public final class IndexStoreDB {
       return body(SymbolOccurrence(occur))
     }
   }
+
+  public func foreachCanonicalSymbolOccurrenceByName(symbolName: String, body: @escaping (SymbolOccurrence) -> Bool) {
+    indexstoredb_for_each_canonical_symbol_occurence_by_name(impl, symbolName) { (occur) -> Bool in
+      body(SymbolOccurrence(occur))
+    }
+  }
+
+  public func findSymbols(matching query: String) -> [SymbolOccurrence] {
+    var symbolNameResults: [String] = []
+    indexstoredb_for_each_symbol_name(impl) { pointer in
+      let str = String(cString: pointer)
+      if str.range(of: query) != nil {
+        symbolNameResults.append(str)
+      }
+      return true
+    }
+    var symbolOccurenceResults: [SymbolOccurrence] = []
+    symbolNameResults.forEach { (symbolName) in
+      foreachCanonicalSymbolOccurrenceByName(symbolName: symbolName) {occurence in
+        if !occurence.location.isSystem {
+          symbolOccurenceResults.append(occurence)
+        }
+        return true
+      }
+    }
+    return symbolOccurenceResults
+  }
 }
 
 public struct SymbolRole: OptionSet {
--- a/Sources/IndexStoreDB/IndexStoreDB.swift
+++ b/Sources/IndexStoreDB/IndexStoreDB.swift
@@ -81,29 +81,45 @@ public final class IndexStoreDB {
     }
   }
 
-  public func foreachCanonicalSymbolOccurrenceByName(symbolName: String, body: @escaping (SymbolOccurrence) -> Bool) {
+  public func forEachCanonicalSymbolOccurrenceByName(symbolName: String, body: @escaping (SymbolOccurrence) -> Bool) {
     indexstoredb_for_each_canonical_symbol_occurence_by_name(impl, symbolName) { (occur) -> Bool in
       body(SymbolOccurrence(occur))
     }
   }
 
+  public func forEachCanonicalSymbolOccurrence(
+    containing pattern: String,
+    anchorStart: Bool,
+    anchorEnd: Bool,
+    subsequence: Bool,
+    ignoreCase: Bool,
+    body: @escaping (SymbolOccurrence) -> Bool
+  ){
+    indexstoredb_for_each_canonical_symbol_occurence_containing_pattern(
+      impl,
+      pattern,
+      anchorStart,
+      anchorEnd,
+      subsequence,
+      ignoreCase
+    ) { occur in
+      body(SymbolOccurrence(occur))
+    }
+  }
+
   public func findSymbols(matching query: String) -> [SymbolOccurrence] {
-    var symbolNameResults: [String] = []
-    indexstoredb_for_each_symbol_name(impl) { pointer in
-      let str = String(cString: pointer)
-      if str.range(of: query) != nil {
-        symbolNameResults.append(str)
+    var symbolOccurenceResults: [SymbolOccurrence] = []
+    forEachCanonicalSymbolOccurrence(
+      containing: query,
+      anchorStart: false,
+      anchorEnd: false,
+      subsequence: true,
+      ignoreCase: true
+    ) { (occurence) -> Bool in
+      if !occurence.location.isSystem {
+        symbolOccurenceResults.append(occurence)
       }
       return true
-    }
-    var symbolOccurenceResults: [SymbolOccurrence] = []
-    symbolNameResults.forEach { (symbolName) in
-      foreachCanonicalSymbolOccurrenceByName(symbolName: symbolName) {occurence in
-        if !occurence.location.isSystem {
-          symbolOccurenceResults.append(occurence)
-        }
-        return true
-      }
     }
     return symbolOccurenceResults
   }
--- a/Sources/IndexStoreDB/IndexStoreDB.swift
+++ b/Sources/IndexStoreDB/IndexStoreDB.swift
@@ -81,21 +81,21 @@ public final class IndexStoreDB {
     }
   }
 
-  public func forEachCanonicalSymbolOccurrenceByName(symbolName: String, body: @escaping (SymbolOccurrence) -> Bool) {
-    indexstoredb_for_each_canonical_symbol_occurence_by_name(impl, symbolName) { (occur) -> Bool in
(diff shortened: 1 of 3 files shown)


literalpie:
I've created [a branch](https://github.com/literalpie/sourcekit-lsp/tree/workspace-symbols) in the sourcekit-lsp project with that side of this change.

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -80,6 +80,33 @@ public final class IndexStoreDB {
...
+    indexstoredb_for_each_canonical_symbol_occurence_by_name(impl, symbolName) { (occur) -> Bool in
+      body(SymbolOccurrence(occur))
+    }
+  }
+
+  public func findSymbols(matching query: String) -> [SymbolOccurrence] {
+    var symbolNameResults: [String] = []
+    indexstoredb_for_each_symbol_name(impl) { pointer in
+      let str = String(cString: pointer)
+      if str.range(of: query) != nil {
```
literalpie: I just realized I should make this case insensitive. I will make the change tonight

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -80,6 +80,33 @@ public final class IndexStoreDB {
...
     }
   }
+
+  public func foreachCanonicalSymbolOccurrenceByName(symbolName: String, body: @escaping (SymbolOccurrence) -> Bool) {
+    indexstoredb_for_each_canonical_symbol_occurence_by_name(impl, symbolName) { (occur) -> Bool in
+      body(SymbolOccurrence(occur))
+    }
+  }
+
+  public func findSymbols(matching query: String) -> [SymbolOccurrence] {
```
akyrtzi: There's already an indexstoredb function to do a search via a pattern, it is `foreachCanonicalSymbolOccurrenceContainingPattern`. You should expose and use that on the sourcekit-lsp side.

include/CIndexStoreDB/CIndexStoreDB.h
```diff
@@ -85,6 +85,9 @@ typedef _Nullable indexstoredb_indexstore_library_t(^indexstore_library_provider
 /// Returns true to continue.
 typedef bool(^indexstoredb_symbol_occurrence_receiver_t)(_Nonnull indexstoredb_symbol_occurrence_t);
 
+/// Returns true to continue.
+typedef bool(^indexstoredb_for_each_symbol_receiver)(const char *_Nonnull);
```
akyrtzi: Per the naming pattern of other block types and to make it clear this receives a string, not a symbol, this should be named `indexstoredb_symbol_name_receiver`.

lib/CIndexStoreDB/CIndexStoreDB.cpp
```diff
@@ -144,6 +144,37 @@ indexstoredb_symbol_name(indexstoredb_symbol_t symbol) {
...
+/// loops through each symbol in the index and calls the receiver function with each symbol
+/// @param index an IndexStoreDB object which contains the symbols
+/// @param receiver a function to be called for each symbol, the CString of the symbol will be passed in to this function.
+/// The function should return a boolean indicating whether the looping should continue.
+bool
+indexstoredb_for_each_symbol_name(indexstoredb_index_t index, indexstoredb_for_each_symbol_receiver receiver) {
+  // indexSystem has foreachsymbolName.
+  auto obj = (IndexStoreDBObject<std::shared_ptr<IndexSystem>> *)index;
+  return obj->value->foreachSymbolName([&](StringRef ref) -> bool {
+    return receiver(ref.str().c_str());
```
akyrtzi: The number of symbol names is going to be huge in large projects, and this is copying every name string to a `std::string`. I recommend to eliminate the string copy by:
* Have the name receiver receive the StringRef pointer and size
* On the Swift side, construct a String by forming a `UnsafeBufferPointer<UInt8>` from the pointer+size pair and pass it to `String(decoding: textBuffer, as: UTF8.self)`

literalpie: oh, thank you! I was hoping there was a way to make this better. Your other comments are also helpful. I'll work on this as soon as I get a chance.

literalpie: Actually, I don't think I'm going to need to worry about the copying problem anymore if I use the pattern matching method. Is that correct?

akyrtzi: It will be general goodness to expose the other functions as well, since we may need them later on in the future, but it is not a requirement, feel free to just expose the pattern function if you'd like to keep it simple.

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -80,6 +80,49 @@ public final class IndexStoreDB {
       return body(SymbolOccurrence(occur))
     }
   }
+
+  public func forEachCanonicalSymbolOccurrenceByName(symbolName: String, body: @escaping (SymbolOccurrence) -> Bool) {
```
benlangmuir: I suggest the name `forEachCanonicalSymbolOccurrence(byName:body:)`

lib/CIndexStoreDB/CIndexStoreDB.cpp
```diff
@@ -144,6 +144,69 @@ indexstoredb_symbol_name(indexstoredb_symbol_t symbol) {
   return obj->value->getName().c_str();
 }
 
+/// loops through each symbol in the index and calls the receiver function with each symbol
+/// @param index an IndexStoreDB object which contains the symbols
+/// @param receiver a function to be called for each symbol, the CString of the symbol will be passed in to this function.
+/// The function should return a boolean indicating whether the looping should continue.
+bool
```
benlangmuir: The documentation comments should go in the header, not the .cpp file.  Please capitalize the first word as well.  Same for the other functions.

lib/CIndexStoreDB/CIndexStoreDB.cpp
benlangmuir: Let's not expose this function if we're not going to fix the string performance issue.  We can always come back to it later.

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -80,6 +80,49 @@ public final class IndexStoreDB {
...
+      anchorStart,
+      anchorEnd,
+      subsequence,
+      ignoreCase
+    ) { occur in
+      body(SymbolOccurrence(occur))
+    }
+  }
+
+  public func findSymbols(matching query: String) -> [SymbolOccurrence] {
```
benlangmuir: This method seems like it's fairly specific to what you want to do in sourcekit-lsp and the name doesn't make it clear what it's doing (e.g. skipping system occurrences?).  I suggest we remove it from the index API and just call `forEachCanonicalSymbolOccurrence(containing:...)` from `sourcekit-lsp`.

lib/CIndexStoreDB/CIndexStoreDB.cpp
```diff
@@ -144,6 +144,69 @@ indexstoredb_symbol_name(indexstoredb_symbol_t symbol) {
...
+/// @param AnchorEnd when true, symbol names should only be considered matching when the first characters of the symbol name match the pattern.
+/// @param Subsequence when true, symbols will be matched even if the pattern is not matched contiguously.
+/// @param IgnoreCase when true, symbols may be returned even if the case of letters does not match the pattern.
+/// @param receiver a function to be called for each canonical occurence that matches the pattern.
+/// The SymbolOccurenceRef of the symbol will be passed in to this function.
+/// The function should return a boolean indicating whether the looping should continue.
+bool
+indexstoredb_for_each_canonical_symbol_occurence_containing_pattern(
+  indexstoredb_index_t index,
+  const char *_Nonnull Pattern,
```
benlangmuir: For new C++ code, let's stick to lowerCamelCase names in variables.  This applies to the other function parameters as well.

Sources/IndexStoreDB/IndexStoreDB.swift
akyrtzi: It should also return `Bool` and have `@discardableResult` like the other functions. Same with `forEachCanonicalSymbolOccurrence`.

include/CIndexStoreDB/CIndexStoreDB.h
```diff
@@ -159,6 +162,27 @@ indexstoredb_error_get_description(_Nonnull indexstoredb_error_t);
 INDEXSTOREDB_PUBLIC void
 indexstoredb_error_dispose(_Nullable indexstoredb_error_t);
 
+INDEXSTOREDB_PUBLIC bool
+indexstoredb_for_each_symbol_name(_Nonnull indexstoredb_index_t index, _Nonnull indexstoredb_symbol_name_receiver);
+
+INDEXSTOREDB_PUBLIC bool
+indexstoredb_for_each_canonical_symbol_occurence_containing_pattern(
```
benlangmuir: To be consistent with the other symbol names, I suggest the following:
* `indexstoredb_index_symbol_names`
* `indexstoredb_index_canonical_symbol_occurences_containing_pattern`
* `indexstoredb_index_canonical_symbol_occurences_by_name`

We've been using "index_" in the name since these are effectively methods of the index object, and I've been using plurals instead of "for_each".  I'm not wed to these names, but it would be good to be consistent.

literalpie:
I believe I've addressed all of the comments again. Thank you again for the feedback!

akyrtzi:
@swift-ci Please test

akyrtzi:
@swift-ci Please test

PR #30, opened 2019-07-11
Add SymbolRelation class with roles and symbols

literalpie:
Add kind property to Symbol

Add forEachRelation of occurrence function

Based on discussions in [workspace/symbol PR](https://github.com/apple/sourcekit-lsp/pull/118) in apple/sourcekit-lsp.

### DIFF
--- a/Sources/IndexStoreDB/IndexStoreDB.swift
+++ b/Sources/IndexStoreDB/IndexStoreDB.swift
@@ -109,6 +109,17 @@ public final class IndexStoreDB {
 
 }
 
+public final class SymbolRelation {
+  let value: indexstoredb_symbol_relation_t
+
+  public lazy var roles: SymbolRole = SymbolRole(rawValue: indexstoredb_symbol_relation_get_roles(value))
+  public lazy var symbol: Symbol = Symbol(indexstoredb_symbol_relation_get_symbol(value))
+
+  public init(_ value: indexstoredb_symbol_relation_t) {
+    self.value = value
+  }
+}
+
 public struct SymbolRole: OptionSet {
 
   public var rawValue: UInt64
@@ -148,12 +159,45 @@ public struct SymbolRole: OptionSet {
   }
 }
 
+public enum IndexSymbolKind {
+  case unknown
+  case module
+  case namespace
+  case namespaceAlias
+  case macro
+  case `enum`
+  case `struct`
+  case `class`
+  case `protocol`
+  case `extension`
+  case union
+  case `typealias`
+  case function
+  case variable
+  case field
+  case enumConstant
+  case instanceMethod
+  case classMethod
+  case staticMethod
+  case instanceProperty
+  case classProperty
+  case staticProperty
+  case constructor
+  case destructor
+  case conversionFunction
+  case parameter
+  case using
+
+  case commentTag
+}
+
 public final class Symbol {
 
   let value: indexstoredb_symbol_t
 
   public lazy var usr: String = String(cString: indexstoredb_symbol_usr(value))
   public lazy var name: String = String(cString: indexstoredb_symbol_name(value))
+  public lazy var kind: IndexSymbolKind = getSymbolKind(from: indexstoredb_symbol_kind(value))
 
   init(_ value: indexstoredb_symbol_t) {
     self.value = value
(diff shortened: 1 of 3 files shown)


include/CIndexStoreDB/CIndexStoreDB.h
```diff
@@ -201,6 +203,30 @@ indexstoredb_index_canonical_symbol_occurences_containing_pattern(
...
+/// @param occurrence The symbol occurrence that whose relations should be found.
+/// @param applier The function that should be performed on each symbol relation.
+/// The function should return a boolean indicating whether the looping should continue.
+INDEXSTOREDB_PUBLIC bool
+indexstoredb_symbol_occurrence_relations(_Nonnull indexstoredb_symbol_occurrence_t,
+                                         bool(^ _Nonnull applier)(indexstoredb_symbol_relation_t _Nonnull ));
+
+/// Get the SymbolKind of a Symbol
+/// @param symbol The symbol whose kind should be found.
+INDEXSTOREDB_PUBLIC uint8_t
```
literalpie: I think the return type should change, but I don't know to what. What would the best way be to convert from cpp SymbolKind to Swift SymbolKind?

Here I am returning the raw uint8_t value that underlies the cpp type, but that's not helpful to a swift consumer because they don't know how to convert that to something meaningful (it doesn't look like the raw value corresponds to the equivalent Swift enum).

I could use `getSymbolKindString` and a switch statement to return a String, but is there a better way than that?

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -107,6 +107,26 @@ public final class IndexStoreDB {
     }
   }
 
+  @discardableResult public func forEachRelation(
```
literalpie: This function isn't strictly necessary since the relations attribute exists in SymbolOccurrence, but I figured it would be useful to expose so people can iterate through relations more efficiently?

benlangmuir: I think it's good to expose the forEach in addition to the method returning an Array, but it should be a method on the occurrence, not on the index.  The difference between finding relations at a specific occurrence, finding all relations of a symbol, and finding all symbols related to a symbol can be kind of confusing, and I think it will help if we put the global queries on the index, and have the simpler "what are the relations of this particular occurrence" on the occurrence itself.

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -200,6 +222,15 @@ public final class SymbolOccurrence {
   deinit {
     indexstoredb_release(value)
   }
+
+  private func getRelations() -> [SymbolRelation] {
+    var relations: [SymbolRelation] = []
+    indexstoredb_symbol_occurrence_relations(value){ relation in
```
benlangmuir: If we move the forEach onto the occurrence, we can implement this method in terms of that.

include/CIndexStoreDB/CIndexStoreDB.h
benlangmuir: It's a bit of duplication, but we should really have an `indexstoredb_symbol_kind_t` enum that duplicates the symbol kinds from the lower-level `indexstore_symbol_kind_t`, and then also have a pure Swift implementation using swiftier names.  See for example how we exposed `SymbolRoles` and `indexstoredb_symbol_role_t`.

The return type here can then be `indexstoredb_symbol_kind_t `.

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -168,13 +159,91 @@ public struct SymbolRole: OptionSet {
   }
 }
 
+public enum IndexSymbolKindEnum: Int {
```
literalpie: I'm not sure if we actually want an enum, or to mimic an enum with `RawRepresentable` like `SymolRole`.

If we make it an enum, it's cleaner, but we cannot directly use the values from the C enum and risk things getting out of sync.

I've included both approaches here. Let me know which one I should remove

literalpie: I'm calling it `IndexSymbolKind` rather than `SymbolKind` because otherwise there are naming collisions in `SourceKit-LSP`. I cannot find a way to use library name spacing when there's also a class with the same name as the library.

benlangmuir: It should be an enum, but it should use a switch to convert between them, not use the raw value directly.  That way we are using the C value by name.

benlangmuir: > I cannot find a way to use library name spacing when there's also a class with the same name as the library.

Not sure what you mean here: doesn't this only apply to the `IndexStoreDB` class itself (which has no conflict)?

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -109,6 +109,17 @@ public final class IndexStoreDB {
 
 }
 
+public final class SymbolRelation {
+  let value: indexstoredb_symbol_relation_t
+
+  public lazy var roles = SymbolRole(rawValue: indexstoredb_symbol_relation_get_roles(value))
```
benlangmuir: Please add type annotations (`: SymbolRole`) to any class/struct properties.  It makes the API more explicit and can improve compile times.

Sources/IndexStoreDB/IndexStoreDB.swift
benlangmuir: By the way, after that change it won't need to be an `Int` raw value enum anymore

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -148,12 +159,91 @@ public struct SymbolRole: OptionSet {
   }
 }
 
+public enum IndexSymbolKindEnum: Int {
+  case unknown = 0,
```
benlangmuir: Preferred style is each case is declared explicitly:
```
case module
case namespace
...
```

Sources/IndexStoreDB/IndexStoreDB.swift
literalpie: To differentiate this `SymbolKind` from the LanguageServerProtocol `SymbolKind`, I tried doing `IndexStoreDB.SymbolKind` (trying to specify that I want SymbolKind to come from the IndexStoreDB library), but this resulted in an error because it seemed to look for SymbolKind in the `IndexStoreDB` class rather than looking for it in the library.

I think my problem is similar to [this stackoverflow](https://stackoverflow.com/questions/26774101/swift-namespace-conflict)

benlangmuir: Got it.  `IndexSymbolKind` (without `Enum`) seems fine for now.

benlangmuir:
Thanks! I think the type annotation for `public lazy var roles` is the only review comment not addressed at this point.

benlangmuir:
@swift-ci please test

PR #36, opened 2019-08-08
[test] Attempt to fix quoting issue on bot

benlangmuir:
After ninja was updated, started seeing extra quote characters around these paths which interfered with tests.


benlangmuir:
@swift-ci please test

PR #37, opened 2019-08-08
[test] Add syntax for marking test location ending at comment

benlangmuir:
Using a leading '<', we can now specify a location ending at the comment
instead of after it.  For example,

```swift
/*a:start*/abc/*<a:end*/
//         ^  ^
//   a:start  a:end

```

This is useful for forming source ranges.

### DIFF
--- a/Sources/ISDBTestSupport/TestLocationScanner.swift
+++ b/Sources/ISDBTestSupport/TestLocationScanner.swift
@@ -26,6 +26,16 @@ import Foundation
 ///   """, url: myURL)
 /// scanner.result == ["foo:def": TestLocation(url: myURL, line: 1, column: 17)]
 /// ```
+///
+/// ## Special Syntax
+///
+/// If the location starts with `<', it will be the location of the start of the
+/// comment instead of the end. E.g.
+///
+/// ```
+/// /*a*/   // column 6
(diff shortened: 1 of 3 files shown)


benlangmuir:
@swift-ci please test

benlangmuir:
> How about as a follow-up that if a certain pattern is used I can get the range via a name, like .range("blah") resulting in some TestRange?

Yeah, I want to do something along those lines, but I haven't decided what is the best ergonomics for the API yet.  My intention was to wait until we had a few more examples and then play around with the different options.

Usually you don't want a `Range<TestLoc>`, but instead a `Range<T>` where `T` is something you converted from a `TestLoc`, such as a `Location` or `Position` in SourceKit-LSP.  In the index itself I haven't had a use for it yet, since we only store single locations.

Some possibilities:

```swift
Position(testLoc("a")) ..< Position(testLoc("a:end")) // Current
Range<Position>(testLoc("a") ..< testLoc("a:end"))
Range<Position>(testLoc("a").range("a:end"))
Range<Position>(testRange("a", "a:end"))
```

We could also add a convention that "foo" and "foo:end" are the way to spell a range and simplify it to a single named entity:

```swift
Range<Position>(testRange("a"))
```

PR #39, opened 2019-08-09
[tibs] Add -sdk to swift compiles on macOS

benlangmuir:
When compiling with toolchains that do not include the swift stdlib
(e.g. the Xcode toolchain), we need the SDK path to be explicitly passed
as an argument.


benlangmuir:
@swift-ci please test

PR #40, opened 2019-08-10
[tibs] Consolidate uses of Process and fix warnings

benlangmuir:
Fix warnings by using the non-deprecated APIs, and while we're at it
consolidate all the uses into a single place that checks the result and
collects stdout. Inspired by SwiftPM's checkNonZeroExit.


benlangmuir:
@swift-ci please test

PR #42, opened 2019-08-11
IndexStore: correct dllstorage attributes (NFC)

compnerd:
The dllstorage appertains to function and variable (declaration,
applied to definitions).  The usage here was to members of a type
definition which does not make sense.  This has no change as the
annotation is ignored with a warning.


compnerd:
@swift-ci please smoke test

benlangmuir:
@swift-ci please test

PR #43, opened 2019-08-12
Remove unnecessary refcounting and allocation for symbol, occurrence, relation

benlangmuir:
These objects are now valid only during the callbacks that provide them
from the index. Since we already converted them immediately, this didn't
impact the Swift code except to remove the indexstoredb_release calls
for these. I updated the documentation to indicate the new lifetime.

Incidentally, relation was using the make_object machinery even though
it was documented to only live for the callback, so this fixes a leak of
relations.


benlangmuir:
@swift-ci please test

PR #47, opened 2019-08-12
Use include instead of import

gmittert:
Windows doesn't support `#import` and using `#include` instead doesn't
affect this test case.


benlangmuir:
@swift-ci please test

PR #48, opened 2019-08-12
Search for Shared Libraries Properly on Windows

gmittert:
- Shared libs have a .dll extension
- dlls are typically store in bin along side the executables on Windows


gmittert:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #49, opened 2019-08-12
Handle Spaces in Makefile Parsing

gmittert:
In a Makefile, spaces can be escaped by a preceeding backslash. Rather
than using string.split, iterate through the string keeping track of the
previous character.


gmittert:
Rebased, added a test, and moved the Makefile logic into its own file.

@swift-ci please test

PR #50, opened 2019-08-12
Fix Ninja Output for Windows

gmittert:
- Ninja requires that `:`s in filenames be escaped as `$:`
- Explicitly call cmd.exe /C on Windows
- In addition, escape spaces for the command that we pass to cmd.exe

I stole the command line quoting verbatim from the one [I wrote for Foundation.]( https://github.com/apple/swift-corelibs-foundation/blob/c5c35c1a59b0a9a05b2e1ffbf8a7bab0a3e59baa/Foundation/Process.swift#L106)

### DIFF
--- a/Sources/ISDBTibs/TibsBuilder.swift
+++ b/Sources/ISDBTibs/TibsBuilder.swift
@@ -262,23 +262,36 @@ extension TibsBuilder {
   }
 
   public func writeNinjaRules<Output: TextOutputStream>(to stream: inout Output) {
+#if os(Windows)
+    let callCmd = "cmd.exe /C "
+#else
+    let callCmd = ""
+#endif
+    let swiftIndexCommand = callCmd + """
+      \(escapeCommand([toolchain.swiftc.path])) $in $IMPORT_PATHS -module-name $MODULE_NAME \
+       -index-store-path index -index-ignore-system-modules \
+       -output-file-map $OUTPUT_FILE_MAP \
+      -emit-module -emit-module-path $MODULE_PATH -emit-dependencies \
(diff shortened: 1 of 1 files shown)


Sources/ISDBTibs/TibsBuilder.swift
```diff
@@ -276,23 +352,36 @@ extension TibsBuilder {
   }
 
   public func writeNinjaRules<Output: TextOutputStream>(to stream: inout Output) {
+#if os(Windows)
+    let callCmd = "cmd.exe /C "
+#else
+    let callCmd = ""
+#endif
+    let swiftIndexCommand = callCmd + """
+\(escapeCommand([toolchain.swiftc.path])) $in $IMPORT_PATHS -module-name $MODULE_NAME \
```
benlangmuir: Indentation

Sources/ISDBTibs/TibsBuilder.swift
```diff
@@ -257,6 +257,82 @@ extension TibsBuilder {
     return result
   }
 
+#if os(Windows)
+  private func quoteWindowsCommandLine(_ commandLine: [String]) -> String {
```
benlangmuir: This and the function that calls it look independent of `self`.  I suggest putting them at the bottom of the file, outside the type.

Sources/ISDBTibs/TibsBuilder.swift
gmittert: Oh neat, I didn't realize that indentation in multilined strings looked at the indent of the first line.

gmittert:
Moved functions out of the class, rebased, and fixed indentation

@swift-ci please test

PR #53, opened 2019-08-17
[5.1] [test] Bump minimum delay for incremental build after change

benlangmuir:
We saw a test for incremental build fail once in CI. I'm not sure what
exactly the behaviour on that bot is, but in Docker I can reproduce a
timestamp resolution of a bit more than 10 ms. So bump the minimum to 20
ms to workaround.

rdar://54419544


benlangmuir:
@swift-ci please test

PR #54, opened 2019-08-23
[test] Fix minimum sleep change

benlangmuir:
My previous change to add a minimum sleep time when testing modified
sources was accidentally broken because I had `if usec > minimum` - that
was only intended to silence the warning, not disable the sleep itself.

When I tested this originally I must have gotten lucky; it seems to fail
1/20 or so in Docker for me. After this commit, I haven't seen any
failures after several hundred runs.

[SR-11356](https://bugs.swift.org/browse/SR-11356)
rdar://54612870


benlangmuir:
@swift-ci please test

PR #55, opened 2019-10-02
Fix crash when printing UnsafeMutableRawPointer

benlangmuir:
We accidentally added a conformance to UnsafeMutablePointer, since the
error is actually a C typedef of a void pointer. Luckily we don't care
about the conformance, only the concrete implementation here.


benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test macOS

PR #56, opened 2019-10-03
Update Package.swift to build on Windows

gmittert:
- To build llvmSupport, we need to define LLVM_ON_WIN32
- To enable cxxsettings, we need swift-tools-version:5.0
- Windows needs c++14 enabled to successfully build the project. Since
  LLVM has moved to c++14, this seems like a reasonable thing to do
  anyways.

These changes allow me to build indexstore-db on Windows using swiftpm.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:4.2
+// swift-tools-version:5.0
 
 import PackageDescription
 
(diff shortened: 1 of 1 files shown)


benlangmuir:
https://github.com/apple/swift-package-manager/pull/2362

@swift-ci please test

Package.swift
```diff
@@ -88,8 +88,12 @@ let package = Package(
     .target(
       name: "IndexStoreDB_LLVMSupport",
       dependencies: [],
-      path: "lib/LLVMSupport"),
+      path: "lib/LLVMSupport",
+      cxxSettings: [
+        .define("LLVM_ON_WIN32", .when(platforms: [.windows])),
```
benlangmuir: Does this propagate to dependent targets, or is that not needed? E.g. when the other targets `#include` llvm headers from this target, will they also need `LLVM_ON_WIN32`?

benlangmuir:
Looks like we also need more changes to match that SwiftPM PR: 
```
Platform.swift:19:5: error: switch must be exhaustive
14:21:09     switch self {
``

akyrtzi:
Is this PR still relevant?

gmittert:
Swiftpm still needs a bit of work so I'll close this in the mean time. Something like this will need to be eventually merged for swiftpm windows support.

PR #60, opened 2019-12-16
Pass through waitUntilDoneInitializing option in Index init

benlangmuir:
When initializing the database, allow waiting for the initial scan to
finish. The existing behaviour is useful for tools like an editor where
you want to return as quickly as possible and don't need to wait for the
data to populate, but for something like a command-line tool making a
query where it expects the data to be ready we needed a way to finish
loading before continuing.


benlangmuir:
@swift-ci please test

PR #61, opened 2019-12-26
Android: add support in TibsToolchain, LLVM config, and fix lmdb ifdefs

buttaface:
I used this pull to build and test natively on Android in [the Termux app](https://termux.com/), all tests passing.

### DIFF
--- a/Sources/ISDBTibs/TibsToolchain.swift
+++ b/Sources/ISDBTibs/TibsToolchain.swift
@@ -55,7 +55,8 @@ public final class TibsToolchain {
   }()
 
   public private(set) lazy var clangHasIndexSupport: Bool = {
-    clangVersionOutput.starts(with: "Apple") || clangVersionOutput.contains("swift-clang")
+    clangVersionOutput.starts(with: "Apple") || clangVersionOutput.contains("swift-clang") ||
+    clangVersionOutput.contains("apple/llvm-project")
   }()
 
   public private(set) lazy var ninjaVersion: (Int, Int, Int) = {
@@ -189,6 +190,8 @@ public func findTool(name: String) -> URL? {
     .appendingPathComponent("system32")
     .appendingPathComponent("where.exe")
   let cmd = [wherePath, name]
+#elseif os(Android)
+  let cmd = ["/system/bin/which", name]
 #else
   let cmd = ["/usr/bin/which", name]
 #endif
--- a/include/llvm/Config/config.h
+++ b/include/llvm/Config/config.h
@@ -17,7 +17,9 @@
 #define LLVM_ENABLE_CRASH_DUMPS 0
 
 /* Define to 1 if you have the `backtrace' function. */
+#if !defined(__ANDROID__)
 #define HAVE_BACKTRACE TRUE
+#endif
 
 #define BACKTRACE_HEADER <execinfo.h>
 
@@ -86,7 +88,9 @@
 /* #undef HAVE_FFI_H */
 
 /* Define to 1 if you have the `futimens' function. */
-/* #undef HAVE_FUTIMENS */
+#if defined(__ANDROID__)
+#define HAVE_FUTIMENS 1
+#endif
 
 /* Define to 1 if you have the `futimes' function. */
 #define HAVE_FUTIMES 1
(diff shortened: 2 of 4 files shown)


Sources/ISDBTibs/TibsToolchain.swift
```diff
@@ -55,7 +55,7 @@ public final class TibsToolchain {
   }()
 
   public private(set) lazy var clangHasIndexSupport: Bool = {
-    clangVersionOutput.starts(with: "Apple") || clangVersionOutput.contains("swift-clang")
+    clangVersionOutput.starts(with: "Apple") || clangVersionOutput.contains("apple/llvm-project")
```
buttaface: The clang from Swift's forked repo doesn't put out either of these on Android. I checked on linux, and neither the latest 5.1.3 release for Ubuntu nor a snapshot release from master say "swift-clang", so I'm not sure how the CI doesn't show this warning I got on Android before I modified this:
```
Test Case 'IndexTests.testMixedLangTarget' started at 2019-12-26 02:07:33.359
warning: skipping test because '/data/data/com.termux/files/home/src/swift/build/Ninja-Release/toolchain-android-aarch64/usr/bin/clang' does not have indexstore support; use swift-clang
Test Case 'IndexTests.testMixedLangTarget' passed (0.153 seconds)
```
The 5.1.3 release had neither, while the snapshot release had this new string, probably ever since the switch to the LLVM monorepo. Maybe the CI is picking up some other clang installed in the system?

I'm unsure if the 5.2 release will have this string or not, so this change is tentative for at least the master branch.

include/llvm/Config/config.h
```diff
@@ -17,7 +17,9 @@
 #define LLVM_ENABLE_CRASH_DUMPS 0
 
 /* Define to 1 if you have the `backtrace' function. */
+#if !defined(__ANDROID__)
 #define HAVE_BACKTRACE TRUE
+#endif
```
buttaface: Pretty much the same as [an earlier patch for llbuild](https://github.com/apple/swift-llbuild/pull/553/files#diff-2675b3ae9211823220c66e10862f930c).

lib/Database/lmdb/mdb.c
```diff
@@ -144,7 +144,7 @@ typedef SSIZE_T	ssize_t;
 #include <unistd.h>
 #endif
 
-#if defined(__sun) || defined(ANDROID)
+#if defined(__sun)
```
buttaface: `posix_memalign` was [added to Bionic with 4.2, ie API 17](https://android.googlesource.com/platform/bionic/+/master/libc/include/stdlib.h#80), a long time ago, which is just one higher than the current minimum supported API of 16 in the NDK, so I think it's safe to assume it's there.

lib/Database/lmdb/mdb.c
```diff
@@ -164,7 +164,7 @@ typedef SSIZE_T	ssize_t;
 # define MDB_USE_SYSV_SEM	1
 # endif
 # define MDB_FDATASYNC		fsync
-#elif defined(ANDROID)
+#elif defined(__ANDROID__)
```
buttaface: You need the underscores, [as can be seen elsewhere in this repo](https://github.com/apple/indexstore-db/blob/master/lib/LLVMSupport/Support/Unix/Threading.inc#L96).

Sources/ISDBTibs/TibsToolchain.swift
benlangmuir: Good catch! The test is being skipped on our Linux CI right now for the same reason.  We should probably make it mandatory at least on CI, but that's a separate issue.  I believe we will see

5.1 and earlier: "swift-clang"
5.2 and later: "apple/llvm-project"

We should change this code to check for both cases, at least for now.

lib/Database/lmdb/mdb.c
benlangmuir: This change should be upstreamed.  I see the upstream fixed the `ANDROID` vs. `__ANDROID__` issue, but this code is still disable on Android: https://github.com/LMDB/lmdb/blob/c3e6b4209eed13af4a3670e5f04f42169c08e5c6/libraries/liblmdb/mdb.c#L147

Is this change *necessary* to get Android working, or just an incremental improvement?  If it's necessary, we could take this change immediately.  If not, my preference would be to get this change in upstream and then update our version of lmdb.  CC @akyrtzi in case he feels differently.

benlangmuir:
BTW, sorry for the slow response to this PR. Everyone who usually reviews these changes was on holidays.

lib/Database/lmdb/mdb.c
buttaface: No, it's not necessary, I just went ahead and updated it, removed now.

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

buttaface:
Made the changes you asked for.

>the slow response to this PR. Everyone who usually reviews these changes was on holidays.

No problem, I figured as much.

PR #62, opened 2020-01-18
Collect the newly discovered records and process them outside of the libIndexStore callback

akyrtzi:
This is because processing populates the database and C++ exceptions can be thrown; libIndexStore builds with -fno-exceptions so we cannot be throwing C++ exceptions from inside its frames.

rdar://55930012


akyrtzi:
@swift-ci Please test

PR #63, opened 2020-01-23
[libindexstore] Remove indexstore_unit_event_get_modification_time

benlangmuir:
This was removed from the implementation. While this is an ABI break,
it's one that we may be able to accept since we aren't using this
function for anything in practice.

rdar://58836941


benlangmuir:
@swift-ci please test

PR #65, opened 2020-01-31
Create the index store path if it does not exist

benlangmuir:
We already do this for the database, so add the store as well. This allows you to open the index before it has been populated.


benlangmuir:
@swift-ci please test

PR #68, opened 2020-02-06
Add mainFilesContainingFile and IndexDelegate

benlangmuir:
Expose the mainFile query to lookup the main files that include a given file, which will be used by sourcekit-lsp when looking up header build settings. Also, expose the index delegate processing events so that a client can track ongoing work and see when the index finishes updating.


benlangmuir:
@swift-ci please test

Sources/ISDBTestSupport/ForwardingIndexDelegate.swift
```diff
@@ -0,0 +1,33 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
```
akyrtzi: `2020` for new files 😉

benlangmuir: I did the indexstore-db side of this in December. I just sat on the changes until I could test them thoroughly from the sourcekit-lsp side.  I can still update the dates if you think it's correct :shrug:  Wish we could just remove the dates from the headers.  No one is consistent about keeping them updated in existing files.

akyrtzi: It's fine, I don't really care much.

PR #71, opened 2020-02-23
[lmdb] Update LMDB to to latest version from upstream

akyrtzi:
A summary of the more interesting changes:

* `mdb_env` flag `MDB_PREVMETA` has been renamed to `MDB_PREVSNAPSHOT` and support for it has improved. For a hint on how it can be useful see http://www.openldap.org/its/index.cgi/Software%20Enhancements?id=8704 :

		Generally, running LMDB in NOSYNC mode means all integrity guarantees are gone,
		in even of a system crash. However, if some other thread runs mdb_env_sync
		periodically it's possible that the older DB state is still consistent, even if
		slightly out of date. I have always intended to add an env flag for using the
		older meta page, to potentially access a DB even after the latest txn snapshot
		was corrupted.

AFAICT we could utilize this by periodically calling `mdb_env_sync` and after a crash open the database using `MDB_PREVSNAPSHOT`, instead of re-creating it from scratch.

* The android `ifdef`s have been corrected in upstream as well, so we are back in-sync.

* Fix for rare and difficult to reproduce crashes, on `mdb_env_close` or on writes after `mdb_env_set_mapsize`, when using `MDB_WRITEMAP`. These are issues we have seen as well.
See more details here https://www.openldap.net/its/index.cgi/Software%20Bugs?id=9007

* Fix for potential memory leak, see http://www.openldap.org/lists/openldap-bugs/202001/msg00023.html

### DIFF
--- a/lib/Database/lmdb/lmdb.h
+++ b/lib/Database/lmdb/lmdb.h
@@ -136,7 +136,7 @@
  *
  *	@author	Howard Chu, Symas Corporation.
  *
- *	@copyright Copyright 2011-2018 Howard Chu, Symas Corp. All rights reserved.
+ *	@copyright Copyright 2011-2020 Howard Chu, Symas Corp. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted only as authorized by the OpenLDAP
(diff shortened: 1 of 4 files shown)


akyrtzi:
@swift-ci Please test

PR #73, opened 2020-03-13
Make sure Swift symbol occurrences are associated with the module name they belong to

akyrtzi:


akyrtzi:
@swift-ci Please test

PR #74, opened 2020-03-13
Expose the name of every symbol in the index 

emixb:
There are scenarios where one wants to iterate through the list of all the symbols in the index. Such functionality exists in the underlying lower-level library. 

This PR makes that functionality available to `indexstore-db` users.

### DIFF
--- a/Sources/IndexStoreDB/IndexStoreDB.swift
+++ b/Sources/IndexStoreDB/IndexStoreDB.swift
@@ -199,11 +199,11 @@ public final class IndexStoreDB {
     }
   }
 
-  /// Returns a set with every symbol in the index.
-  public func allSymbolNames() -> Set<String> {
-    var result: Set<String> = []
+  /// Returns an array with every symbol name in the index.
+  public func allSymbolNames() -> [String] {
+    var result: [String] = []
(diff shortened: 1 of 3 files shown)


Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -187,6 +187,27 @@ public final class IndexStoreDB {
...
+  /// continue iterating.
+  @discardableResult
+  public func forEachSymbolName(body: @escaping (String) -> Bool) -> Bool {
+    return indexstoredb_index_symbol_names(impl) { name in
+      body(String(cString: name))
+    }
+  }
+
+  /// Returns a set with every symbol in the index.
+  public func allSymbolNames() -> Set<String> {
```
akyrtzi: This doesn't need to be a `Set`, `indexstoredb_index_symbol_names` is going to return unique strings.
I'd recommend to change it to array.

emixb: Done!

akyrtzi:
@swift-ci test and merge

akyrtzi:
@swift-ci Please test

PR #78, opened 2020-03-25
build: allow building with CMake without symlinks

compnerd:
Windows can still be finicky with symbolic links.  This adjusts the
include paths to allow building without relying on them.

Additionally, make the include path for `indexstore` public so that
users of it will implicitly get its headers.  Lastly, adjust the linking
for CIndexStore to use `PUBLIC` for `indexstore` to allow the header
include paths to be transitively included.  This mirrors the behaviour
of swift-package-manager.

This allows IndexStoreDB to be built on Windows again.


benlangmuir:
@swift-ci please test

compnerd:
Anything else left for this?

PR #79, opened 2020-04-01
[tibs] Detect clang index support using -help instead of --version

benlangmuir:
The previous check for apple-clang, swift-clang, etc. was failing on
released toolchains, since they don't contain the repo name, and on
clones of the repo that didn't match the pattern.

Check clang -help for index store support. It would be better to check
that `clang -index-store-path /dev/null --version` does not produce an
error, but unfortunately older versions of clang accepted
`-index-store-path` due to the existince of a `-i` option that has since
been removed. While we could check a full compile command, I have not
found a robust command that detects index support without also doing I/O
in the index directory when it succeeds. To avoid I/O, we check -help.

### DIFF
--- a/Sources/ISDBTibs/TibsToolchain.swift
+++ b/Sources/ISDBTibs/TibsToolchain.swift
@@ -50,13 +50,27 @@ public final class TibsToolchain {
   public static let dylibExt = "so"
 #endif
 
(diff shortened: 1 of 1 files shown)


benlangmuir:
@swift-ci please test

PR #80, opened 2020-04-07
[build-script] Add --parallel to supress output on success

benlangmuir:
It's slightly faster and it helps supress the test output when tests
succeed (which avoids false positives in Jenkins' error regexes).


benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #81, opened 2020-04-10
[test] Workaround an issue with createDirectory on Linux

benlangmuir:
Looks like createDirectory can fail spuriously if run in parallel with a
common parent path missing. Will file a bug on corelibs-foundation, but
for now workaround to avoid test failures.

rdar://61545973


benlangmuir:
@swift-ci please test

PR #83, opened 2020-04-16
[Index] Provide capability for the client to explicitly specify the list of output unit files for the datastore

akyrtzi:
The benefits of this approach is that it allows the datastore to ignore leftover unit files so it can both improve performance and correctness by not taking into account stale data.

rdar://61899059

### DIFF
--- a/lib/Index/IndexDatastore.cpp
+++ b/lib/Index/IndexDatastore.cpp
@@ -116,7 +116,7 @@ class StoreUnitRepo : public std::enable_shared_from_this<StoreUnitRepo> {
   mutable llvm::sys::Mutex StateMtx;
   std::unordered_map<IDCode, std::shared_ptr<UnitMonitor>> UnitMonitorsByCode;
 
-  llvm::StringSet<> ExplicitOutputUnitsSet;
+  std::unordered_set<db::IDCode> ExplicitOutputUnitsSet;
 
 public:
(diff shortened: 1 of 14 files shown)


akyrtzi:
@swift-ci Please test

akyrtzi:
@swift-ci Please test

akyrtzi:
@swift-ci Please test Windows platform

lib/Index/IndexDatastore.cpp
```diff
@@ -108,12 +116,16 @@ class StoreUnitRepo : public std::enable_shared_from_this<StoreUnitRepo> {
   mutable llvm::sys::Mutex StateMtx;
   std::unordered_map<IDCode, std::shared_ptr<UnitMonitor>> UnitMonitorsByCode;
 
+  llvm::StringSet<> ExplicitOutputUnitsSet;
```
benlangmuir, line 119: Does this need to be the names, or could it be the hashes?  This is going to be a really big set if it stores string data.

akyrtzi: Good point, will do that 👍

akyrtzi:
@swift-ci Please test

akyrtzi:
@swift-ci test linux platform

akyrtzi:
@swift-ci test linux platform

PR #85, opened 2020-04-21
[5.3] Reduce memory usage in onFSEvent callback

benlangmuir:
Cherry-pick #84 and #86 to 5.3

----

Store the path of the file triggering an out-of-date check once in the
array of checks, rather than once for every unit. For headers included
by many unit files this was causing us to use several GB just for this
single std::vector.

rdar://53835770


benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

PR #86, opened 2020-04-21
Additional memory savings for onFSEvent

benlangmuir:
Argyrios suggested switching to std::string, since CanonicalFilePath
contains a SmallString<128>.

rdar://53835770


benlangmuir:
@swift-ci please test

PR #88, opened 2020-04-22
[IndexDatastore] Provide an option to disable the file-out-of-date notification mechanism

akyrtzi:
This mechanism is disabled by default for the CIndexStoreDB API because the out-of-date notifications are not exposed at all, so the mechanism is taking CPU & memory unnecessarily.

### DIFF
--- a/include/IndexStoreDB/Index/IndexSystem.h
+++ b/include/IndexStoreDB/Index/IndexSystem.h
@@ -48,6 +48,7 @@ public:
                                              std::shared_ptr<IndexSystemDelegate> Delegate,
                                              bool useExplicitOutputUnits,
                                              bool readonly,
+                                             bool enableOutOfDateFileWatching,
                                              bool listenToUnitEvents,
                                              Optional<size_t> initialDBSize,
                                              std::string &Error);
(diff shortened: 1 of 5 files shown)


akyrtzi:
@swift-ci Please test

akyrtzi:
@swift-ci test linux platform

include/IndexStoreDB/Index/IndexSystem.h
```diff
@@ -47,6 +47,7 @@ class INDEXSTOREDB_EXPORT IndexSystem {
                                              std::shared_ptr<IndexStoreLibraryProvider> storeLibProvider,
                                              std::shared_ptr<IndexSystemDelegate> Delegate,
                                              bool useExplicitOutputUnits,
+                                             bool disableOutOfDateFileTriggers,
```
benlangmuir, line 50: This option should align with `listenToUnitEvents`, since they are both related to file watching. I suggest putting these arguments next to each other in the list and rewording it so the new option is positive instead of negative.  I'm less concerned about the specific name, but it should enable something not disable it.

akyrtzi:
@swift-ci Please test

PR #89, opened 2020-04-23
[IndexDataStore] Reduce logging priority for missing units when in explicit-out-units mode

akyrtzi:
It is normal to setup the list of units before the data is generated.


akyrtzi:
@swift-ci Please test

PR #90, opened 2020-04-23
Implicit use of <cstdint.h> fails with Clang 10

tachoknight:
Discovered this while building under Linux


benlangmuir:
@swift-ci please test

PR #95, opened 2020-04-30
[build-script-helper] Always do a clean build, to workaround an incremental build bug in swiftpm

akyrtzi:


akyrtzi:
@swift-ci Please test

PR #97, opened 2020-05-01
build: add CIndexStoreDB to the interface includes

compnerd:
This include path is needed to build SourceKit-LSP.


benlangmuir:
Could we change IndexStoreDB to use `@_implementationOnly import CIndexStoreDB` instead?

compnerd:
My opinion on that: that’s way better.  I can try to see if I can get that to work.

benlangmuir:
@swift-ci please test

PR #98, opened 2020-05-08
[5.3] [tibs] Detect clang index support using -help instead of --version

benlangmuir:
Cherry-pick #79 to 5.3

* Explanation: The way our unit tests detect "index while building" support was brittle in practice, which could cause tests to fail spuriously with (a) a toolchain built locally, depending on how your forks are setup, or (b) when using a final released toolchain from swift.org (e.g. `5.2-RELEASE`).  The fix is to check for `-index-store-path` in `-help` instead of trying to infer whether the `--version` output comes from a swift-clang fork.
* Scope: Test-only change.
* Risk: Low. This is test-only, and is not tied in with anything else.
* Reviewer: @akyrtzi
* Testing: Regression tests pass.
* Bug: rdar://63023189

### DIFF
--- a/Sources/ISDBTibs/TibsToolchain.swift
+++ b/Sources/ISDBTibs/TibsToolchain.swift
@@ -50,13 +50,27 @@ public final class TibsToolchain {
   public static let dylibExt = "so"
 #endif
 
-  public private(set) lazy var clangVersionOutput: String = {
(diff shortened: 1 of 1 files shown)


benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please nominate

PR #99, opened 2020-05-11
[tsan] Fix a race during processUnitsAndWait()

benlangmuir:
After we added processUnitsAndWait() as a separte entry point that
ultimately called the onFilesChange() method, it created a race on the
DoneInitState, which was accessed directly in the unit event callback
before jumping to the unit processing queue. Simplify that logic and fix
the race by using processUnitsAndWait() during initialization if we
desire waiting, and get rid of DoneInitState.

The previous approach was more flexible about allowing you to wait at an
arbitrary point after creating the IndexDatastore, but in practice we
were not using that ability, and it made it harder to understand the
synchronization required.

### DIFF
--- a/include/IndexStoreDB/Index/IndexSystem.h
+++ b/include/IndexStoreDB/Index/IndexSystem.h
@@ -50,10 +50,9 @@ public:
                                              bool readonly,
                                              bool enableOutOfDateFileWatching,
                                              bool listenToUnitEvents,
+                                             bool waitUntilDoneInitializing,
                                              Optional<size_t> initialDBSize,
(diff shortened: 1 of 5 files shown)


benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test Linux

benlangmuir:
https://github.com/apple/swift-corelibs-foundation/pull/2793
@swift-ci please test Linux

benlangmuir:
https://github.com/apple/swift/pull/31733
@swift-ci please test Linux

PR #101, opened 2020-06-04
build: only link against Foundation on non-Darwin targets

compnerd:
Rely on auto-linking of frameworks on Darwin to get the Foundation
dependency.


benlangmuir:
Closing in favour of a more complete fix: https://github.com/apple/indexstore-db/pull/102

PR #102, opened 2020-06-04
[cmake] Fix linking of Foundation and dispatch on Darwin

benlangmuir:
These come from the SDK and we rely on autolinking. Fixes incorrect
-ldispatch and -lFoundation linker flags.


benlangmuir:
@swift-ci please test

PR #108, opened 2020-07-14
[5.3][Database] Make the database path unique to the specific lmdb wrapper instance

akyrtzi:
This is intended to protect against a crash that can occur if trying to use the same lmdb database path from separate wrapper instances.

rdar://39235399

master: https://github.com/apple/indexstore-db/pull/107


akyrtzi:
@swift-ci Please test

PR #110, opened 2020-09-01
Fix race in index delegate callback 

benlangmuir:
All calls to `processingCompleted` are called from the unit processing queue, but `processingAddedPending` can be called from other queues, including an arbitrary queue in the case of `addUnitOutFilePaths` and `removeUnitOutFilePaths`.

The result is that if there is a session being processed, we might increment and decrement `PendingActions` concurrently. We obsevered an assertion failure where `NumActions` was greater than `NumActions`, which was likely caused by this race.

rdar://68024910

---

Also, switch to posix mutex in lmdb. TSan does not support sysv semaphores, so switch to posix mutex. Also,
disable robust mutexes, since we do not support multi-process access to the database anyway.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:4.2
+// swift-tools-version:5.0
 
 import PackageDescription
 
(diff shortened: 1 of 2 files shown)


benlangmuir:
@swift-ci please test

PR #111, opened 2020-09-07
build: disable POSIX mutexes on Windows

compnerd:
Conditionalize the define to use POSIX mutexes to macOS and Linux
targets.  This would be easier to define as non-POSIX, but the package
description does not have the negative check nor a POSIX check.

This requires bumping the tools version to 5.3 to gain access to
Windows.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.2
 
 import PackageDescription
 
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.3
 
 import PackageDescription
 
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
This cannot be merged right now, I want to let the test finish before I update it; I just realized that this would cause a problem for older versions of spm, I'll add a workaround.

compnerd:
Never mind, I didn't use `.when(platforms: [.windows])`, so this should be fine.

benlangmuir:
> This requires bumping the tools version to 5.3 to gain access to Windows.

We should not do this until 5.3 is released.  Now that you're not using `.windows`, is this still necessary?

compnerd:
Ah, right you are.  Not using `.windows` means that we can get away with 5.2!

benlangmuir:
@swift-ci please test

benlangmuir (approved):

PR #112, opened 2020-09-07
build: bump C++ standard to 14

compnerd:
The Windows C++ runtime requires C++14.  This is handled transparently
when using the `clang-cl` driver, however, when building with s-p-m, the
`clang` driver is used instead.  This requires explicitly bumping
everything to C++14.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -95,5 +95,5 @@ let package = Package(
       path: "lib/LLVMSupport"),
   ],
 
-  cxxLanguageStandard: .cxx11
+  cxxLanguageStandard: .cxx14
 )


compnerd:
@akyrtzi @benlangmuir - I'm not sure if this is acceptable or not.  I couldn't figure out a way to do a build condition on the setting here, which would be another option otherwise.  But, the current MSVC C++ STL headers use C++14 features unconditionally.

benlangmuir (approved):
If the tests pass, it's fine with me.  We would need to change this anyway on all platforms next time we update our llvm support code (it's currently from April 2019, which was a few months before c++14 became mandatory in llvm).

benlangmuir:
@swift-ci please test

PR #113, opened 2020-09-18
[test] Add missing dependency on tibs executable

benlangmuir:
This is a run-time dependency inside the tests.  The missing dependency wasn't generally visible when testing from the command-line unless you tried to build+test specific targets.


benlangmuir:
@swift-ci please test

akyrtzi (approved):

PR #114, opened 2020-09-24
[tsan] Workaround uninstrumented library code

benlangmuir:
We get rare reports from TSan that there is a race between modifying the
UnitEventInfoDeque and its own destructor. This appears to be caused by
uninstrumented code in the indexstore library as well as the underlying
OS file watching library (FSEvents, inotify, etc.), which prevents TSan
from realizing that the event stream callback and the event stream
finalizer are never called concurrently. We workaround by introducing an
unnecessary mutex that we lock around the requisite calls.

rdar://69514830

### DIFF
--- a/include/indexstore/IndexStoreCXX.h
+++ b/include/indexstore/IndexStoreCXX.h
@@ -19,6 +19,7 @@
 #include "llvm/ADT/Optional.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SmallString.h"
+#include "llvm/Support/Mutex.h"
 #include <ctime>
 
 namespace indexstore {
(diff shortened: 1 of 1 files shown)


benlangmuir:
@swift-ci please test

akyrtzi:
I'm wondering if we should just add the mutex unconditionally anyway, instead of diverging the code based on whether sanitizer is enabled or not? What do you think, are there performance concerns?

benlangmuir:
:shrug: it's a bit of overhead in every event callback.  I doubt one more lock makes much difference one way or the other.  My thinking was that this change is purely to workaround the lack of visibility to tsan, so it should only affect tsan builds.  It also helps make it clear in the code that something weird is going on.

akyrtzi (approved):

PR #117, opened 2020-10-18
IndexStoreDB: use `ucrt` instead of `MSVCRT`

compnerd:
There is nothing that requires the use of `MSVCRT`.  Use the modern C
library on Windows, which also helps remove the `visualc` module in the
Swift standard library.


compnerd:
@swift-ci please test

benlangmuir (approved):

PR #121, opened 2020-11-30
ISDBTibs: add extension for executables

compnerd:
Windows uses an extension for executable files.  Support this to ensure
that we can correctly identify the executables in the TibsToolchain on
Windows.


compnerd:
@swift-ci please test

benlangmuir (approved):

PR #126, opened 2021-03-15
Add  `symbols(inFilePath path: String)`

SeanROlszewski:
## Overview
Add a new API which can be used to retrieve a list of symbol definitions or declarations from a file, like so:

```swift
func symbols(inFilePath path: String) -> [Symbol] { /* defined in PR */ }
```

## Usage
```swift
let symbols = Globals.index.symbols(inFilePath: path)
for symbol in symbols {
    print(symbol.name)
}
```

## Test Plan
I added integration tests with the guidance of @benlangmuir. These tests withstood mutation testing and caught meaningful errors during the development process.


benlangmuir:
@swift-ci please test

benlangmuir (approved):

PR #128, opened 2021-05-13
Bug Fix for copy_file_internal

johnno1962:
Noticed this bug in the copy_file_internal code. As bytes are written you need to move through the buffer.


benlangmuir:
This fix looks correct to me, but it should be submitted to upstream llvm and then we should update our copy of the llvm sources using `Utilities/import-llvm`.  

Is this code actually used in indexstore-db?  If this is causing active breakage we could take it ahead of making the upstream fix, but I don't see anywhere that uses this code (no references to copy_file except in TempFile, and no references to TempFile other than its own definition).

johnno1962:
Sorry, don't do llvm submits. It's definitely a bug, up to you whether you act on it.

PR #130, opened 2021-06-28
Allow individual units that require a database file > 1 GB

benlangmuir:
We observed such a large unit in practice, so increase our retry limit
so that it can accomodate such a large unit even if it happens to be the
first one registered with the index and has to perform all the doublings
in a row. The goal of the retry limit is not to keep the DB small, but
to abort in the case of a bug causing unlimited growth.

rdar://77288893


benlangmuir:
@swift-ci please test

akyrtzi (approved):

benlangmuir:
@swift-ci please test macOS

PR #132, opened 2021-09-14
[tibs] Extend error types to include process output

benlangmuir:
When there is a failure, include the stdout + stderr if they can be
decoded. This should help track down some failures we're seeing
sporadically in CI.


benlangmuir:
@swift-ci please test

akyrtzi (approved):

PR #133, opened 2021-09-21
[tibs] Workaround issue with new -emit-modules-separately behaviour

benlangmuir:
We are seeing sporadic test failures after the driver started using
-emit-modules-separately behaviour. For now, workaround by adding -c to
the command to force a per-file compilation. Long term we should figure
out a solution that lets us index and produce swiftmodules without
emitting the unnecessary .o files.

rdar://83215771


benlangmuir:
@swift-ci please test

akyrtzi (approved):

PR #134, opened 2021-10-23
Sync symbol properties with libIndexStore

bnbarham:
libIndexStore has a new SwiftAsync symbol property, add this to the
headers.

Also add the missing ProtocolInterface property in IndexStoreDB.

Resolves rdar://83780222


bnbarham:
Doesn't look like properties are passed through into the Swift API, so I didn't add any tests. Would it be worth adding?

PR #137, opened 2021-12-24
[Database] When opening an existing database make sure to preserve its original size

akyrtzi:
Previously it would get truncated to the stored data, meaning we would start using a full database that needs growing.

### DIFF
--- a/lib/Database/Database.cpp
+++ b/lib/Database/Database.cpp
@@ -173,8 +173,16 @@ retry:
     db->DBEnv = lmdb::env::create();
     db->DBEnv.set_max_dbs(14);
 
-    // Start with 64MB. We'll update with the actual size after we open the database.
-    db->MapSize = initialDBSize.getValueOr(64ULL*1024ULL*1024ULL);
+    uint64_t dbFileSize = 0;
+    if (existingDB) {
+      if (std::error_code ec = llvm::sys::fs::file_size(dbPath + "/data.mdb", dbFileSize)) {
+        LOG_WARN_FUNC("failed reading database file size " << dbPath << "/data.mdb: " << ec.message());
+      }
+    }
+    // Start with 64MB.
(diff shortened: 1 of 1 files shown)


lib/Database/Database.cpp
```diff
@@ -186,7 +194,6 @@ Database::Implementation::create(StringRef path, bool readonly, Optional<size_t>
     // Get actual map size of the database.
     MDB_envinfo envInfo;
     lmdb::env_info(db->DBEnv, &envInfo);
-    db->MapSize = envInfo.me_mapsize;
```
bnbarham, line 197: I don't see any other uses of `envInfo`, so can probably remove this block.

lib/Database/Database.cpp
```diff
@@ -173,8 +173,16 @@ Database::Implementation::create(StringRef path, bool readonly, Optional<size_t>
...
+    uint64_t dbFileSize = 0;
+    if (existingDB) {
+      if (std::error_code ec = llvm::sys::fs::file_size(dbPath + "/data.mdb", dbFileSize)) {
+        LOG_WARN_FUNC("failed reading database file size " << dbPath << "/data.mdb: " << ec.message());
+      }
+    }
+    // Start with 64MB.
+    uint64_t initialSize = initialDBSize.getValueOr(64ULL*1024ULL*1024ULL);
+
+    db->MapSize = std::max(dbFileSize, initialSize);
```
bnbarham, line 185: Would there be any harm in just always using the existing size if there's a database? Ie. could we just initialize `dbFileSize` to `initialDBSize.getValueOr...` and overwrite it if there's an existing database? Assuming `file_size` doesn't set `dbFileSize` unless it succeeds that is.

bnbarham (approved):
LGTM other than the small comments

akyrtzi:
@swift-ci Please test

compnerd:
This seems to have regressed the Windows build: https://ci-external.swift.org/job/oss-swift-windows-toolchain-x86_64-vs2019/579/console

PR #138, opened 2022-01-07
Revert "[Database] When opening an existing database make sure to preserve its original size"

compnerd:
Reverts apple/indexstore-db#137

This breaks on windows: https://ci-external.swift.org/job/oss-swift-windows-toolchain-x86_64-vs2019/579/console


compnerd:
@swift-ci please test

compnerd:
I think I can fix the build on Windows, this is incase we cannot get that merged quickly

compnerd:
#139 should fix the windows build - don't see why it would cause any problems on non-Windows, so that should be preferred over this.

PR #139, opened 2022-01-07
Database: repair the Windows build after #137

compnerd:
Without `NOMINMAX`, `max` is defined as a macro which will then collide
with `std::max` from `<algorithm>`.  This repairs the Windows build.


compnerd:
@swift-ci please test

bnbarham (approved):
You could undef `NOMINMAX` after the include, but otherwise LGTM. Not super concerned about that either.

benlangmuir (approved):

akyrtzi:
Thank you for fixing this! 🙇🏻‍♂️

PR #141, opened 2022-01-11
[ISDBTibs/Makefile] Use `String(contentsOf:encoding:)` which gives the more efficient native String backing

akyrtzi:


akyrtzi:
@swift-ci Please test

benlangmuir (approved):

PR #142, opened 2022-01-14
[Index] Don't check a new unit for out-of-date dependencies, unless it is the initial scan

akyrtzi:
Out-of-date checks should happen for when doing the initial scan and when the file watcher finds that a dependency was modified.
Doing an out-of-date check for a newly created unit can lead to constant attempts at re-indexing an out-of-date unit, like
in the case where a dependency has a timestamp in the future, or if the dependency doesn't exist on disk for some reason.

But though the latter (missing dependency) is a problem, constantly sending out-of-date notifications for the unit is not helpful.

rdar://87127317


akyrtzi:
@swift-ci Please test

lib/Index/IndexDatastore.cpp
```diff
@@ -763,17 +766,17 @@ void StoreUnitRepo::pollForUnitChangesAndWait() {
 
       auto I = knownUnits.find(unitName);
       if (I != knownUnits.end() && I->getValue() != modTime) {
-        events.push_back({IndexStore::UnitEvent::Kind::Modified, unitName.str()});
+        events.push_back({IndexStore::UnitEvent::Kind::Modified, unitName.str(), /*isInitialScan=*/false});
```
benlangmuir, line 769: Should we pass in `isInitialScan` to `pollForUnitChangesAndWait()`?  It is used for the initial scan when `listenToUnitEvents` is false.  Otherwise I think this should have a comment to explain why false would always be correct here.

include/CIndexStoreDB/CIndexStoreDB.h
```diff
@@ -215,6 +217,27 @@ indexstoredb_delegate_event_get_kind(_Nonnull indexstoredb_delegate_event_t);
...
+INDEXSTOREDB_PUBLIC uint64_t
+indexstoredb_delegate_event_get_outofdate_modtime(_Nonnull indexstoredb_delegate_event_t);
+
+/// Valid only if the event kind is \p INDEXSTOREDB_EVENT_UNIT_OUT_OF_DATE, otherwise returns false.
+INDEXSTOREDB_PUBLIC bool
+indexstoredb_delegate_event_get_outofdate_is_synchronous(_Nonnull indexstoredb_delegate_event_t);
+
+/// Valid only if the event kind is \p INDEXSTOREDB_EVENT_UNIT_OUT_OF_DATE, otherwise returns null.
+INDEXSTOREDB_PUBLIC const char * _Nullable
+indexstoredb_delegate_event_get_outofdate_trigger_original_file(_Nonnull indexstoredb_delegate_event_t);
```
benlangmuir, line 235: Doc comment should mention the ownership of the string

```swift
/// The string has the same lifetime as the \c indexstoredb_delegate_event_t.
```

Same for `indexstoredb_delegate_event_get_outofdate_trigger_description`.

include/CIndexStoreDB/CIndexStoreDB.h
```diff
@@ -215,6 +217,27 @@ indexstoredb_delegate_event_get_kind(_Nonnull indexstoredb_delegate_event_t);
 INDEXSTOREDB_PUBLIC
 uint64_t indexstoredb_delegate_event_get_count(_Nonnull indexstoredb_delegate_event_t);
 
+/// Valid only if the event kind is \p INDEXSTOREDB_EVENT_UNIT_OUT_OF_DATE, otherwise returns null.
+INDEXSTOREDB_PUBLIC _Nullable indexstoredb_unit_info_t
+indexstoredb_delegate_event_get_outofdate_unit_info(_Nonnull indexstoredb_delegate_event_t);
```
benlangmuir, line 222: Doc comment should mention the ownership of the unit info -- e.g.
```swift
/// The indexstoredb_unit_info_t pointer has the same lifetime as the \c indexstoredb_delegate_event_t
```

lib/CIndexStoreDB/CIndexStoreDB.cpp
```diff
@@ -95,7 +114,7 @@ indexstoredb_index_create(const char *storePath, const char *databasePath,
   if (auto index =
           IndexSystem::create(storePath, databasePath, libProviderObj, delegate,
                               useExplicitOutputUnits, readonly,
-                              /*enableOutOfDateFileWatching=*/false, listenToUnitEvents, wait,
+                              enableOutOfDateFileWatching, listenToUnitEvents, wait,
```
benlangmuir, line 117: There's a comment about `enableOutOfDateFileWatching` above that is no longer accurate for this part of the code.  If the performance considerations are still relevant, they should probably be mentioned in the doc comment for the `enableOutOfDateFileWatching` parameter you added to `IndexStoreDB.init`.

Tests/IndexStoreDBTests/IndexTests.swift
```diff
@@ -253,6 +253,73 @@ final class IndexTests: XCTestCase {
     XCTAssertEqual(delegate.completed, 3)
   }
 
+  func testOutOfDateEvent() throws {
```
benlangmuir, line 256: Need to update Linux tests.  I mentioned this to Alex for SourceKit-LSP, but we should consider switching to rely on automatic test discovery on Linux now that it's been around for a while.  We may need to bump our swift tools version in the package for that, not sure.

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -49,6 +49,7 @@ public final class IndexStoreDB {
     useExplicitOutputUnits: Bool = false,
     waitUntilDoneInitializing wait: Bool = false,
     readonly: Bool = false,
+    enableOutOfDateFileWatching: Bool = false,
```
benlangmuir, line 52: Please add the new parameter to the doc comment.

Tests/IndexStoreDBTests/IndexTests.swift
```diff
@@ -253,6 +253,73 @@ final class IndexTests: XCTestCase {
...
+      let outOfDateModTime: UInt64
+      let triggerHintFile: String
+      let triggerHintDescription: String
+      let synchronous: Bool
+    }
+
+    class Delegate: IndexDelegate {
+      let queue: DispatchQueue = DispatchQueue(label: "testDelegate mutex")
+      private var _outOfDateInfo: OutOfDateInfo?
+      var outofDateInfo: OutOfDateInfo? { queue.sync { _outOfDateInfo } }
```
bnbarham, line 268: Nit: `outofDateInfo` vs `outOfDateInfo`

lib/Index/IndexDatastore.cpp
akyrtzi: Good point, added as parameter in the APIs.

include/CIndexStoreDB/CIndexStoreDB.h
akyrtzi: Done.

include/CIndexStoreDB/CIndexStoreDB.h
akyrtzi: Done.

lib/CIndexStoreDB/CIndexStoreDB.cpp
akyrtzi: Done.

Tests/IndexStoreDBTests/IndexTests.swift
akyrtzi: Good catch! We missed updating for other tests as well. Done.

Sources/IndexStoreDB/IndexStoreDB.swift
akyrtzi: Done.

Tests/IndexStoreDBTests/IndexTests.swift
akyrtzi: Done.

akyrtzi:
@swift-ci Please test

Tests/IndexStoreDBTests/XCTestManifests.swift
```diff
@@ -22,9 +22,14 @@ extension IndexTests {
...
         ("testEditsSimple", testEditsSimple),
         ("testExplicitOutputUnits", testExplicitOutputUnits),
+        ("testFilesIncludes", testFilesIncludes),
         ("testMainFilesContainingFile", testMainFilesContainingFile),
         ("testMixedLangTarget", testMixedLangTarget),
+        ("testOutOfDateEvent", testOutOfDateEvent),
+        ("testProperties", testProperties),
         ("testSwiftModules", testSwiftModules),
+        ("testSymbolsInFileC", testSymbolsInFileC),
+        ("testSymbolsInFileSwift", testSymbolsInFileSwift),
```
benlangmuir, line 32: Guess we were a little behind 😅

benlangmuir (approved):

bnbarham (approved):

Tests/IndexStoreDBTests/XCTestManifests.swift
bnbarham: Some of these are probably mine, sorry, didn't know it was a thing 😅

PR #148, opened 2022-05-13
Support for hermetic indexing information

DavidGoldman:
Make use of the support added in https://github.com/apple/swift/pull/42394 and https://github.com/apple/llvm-project/pull/4267 to create the index store using `indexstore_store_create_with_options` in order to allow path remappings to be used. This lets us remap hermetic index data to the equivalents on the local machine.


include/CIndexStoreDB/CIndexStoreDB.h
```diff
@@ -49,6 +49,7 @@
 INDEXSTOREDB_BEGIN_DECLS
 
 typedef void *indexstoredb_object_t;
+typedef void *indexstoredb_creation_options_t;  // Not a indexstoredb_object_t since it has a separate dispose.
```
bnbarham, line 52: Nitpick: I'd move this down with the others and I'm not sure the comment is really needed - it's not a `indexstoredb_object_t` because it ... isn't one. Eg. same as `indexstoredb_symbol_t`, `indexstoredb_symbol_occurrence_t`, etc.

include/IndexStoreDB/Index/IndexSystem.h
```diff
@@ -15,6 +15,7 @@
 
 #include "IndexStoreDB/Support/LLVM.h"
 #include "IndexStoreDB/Support/Visibility.h"
+#include "indexstore/IndexStoreCXX.h"
```
bnbarham, line 18: Can just forward declare `indexstore::IndexStoreCreationOptions` rather than including the header.

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -52,7 +53,8 @@ public final class IndexStoreDB {
     waitUntilDoneInitializing wait: Bool = false,
     readonly: Bool = false,
     enableOutOfDateFileWatching: Bool = false,
-    listenToUnitEvents: Bool = true
+    listenToUnitEvents: Bool = true,
+    prefixMappings: [String] = []
```
bnbarham, line 57: I think it'd be nice to give this a proper API rather than just taking strings. Eg. `[PathMapping]` with a new `PathMapping` type.

include/indexstore/IndexStoreCXX.h
```diff
@@ -42,6 +44,28 @@ class IndexStoreLibrary {
   const indexstore_functions_t &api() const { return functions; }
 };
 
+class IndexStoreCreationOptions {
+  std::vector<std::pair<std::string, std::string>> prefixMap;
+
+public:
+  IndexStoreCreationOptions() {}
+
+  void addPrefixMapping(const char *orig, const char *remapped) {
```
bnbarham, line 53: Could take `StringRef` here

include/indexstore/IndexStoreCXX.h
```diff
@@ -117,10 +141,22 @@ class IndexStore {
...
+    // prefix mapping.
+    //
+    // TODO: Should we error if the API is unavailable but a mapping is given?
+    if (api().store_create_with_options) {
+      indexstore_creation_options_t c_options = options->createOptions(api());
+      obj = api().store_create_with_options(buf.c_str(), c_options, &c_err);
+      api().creation_options_dispose(c_options);
+    }
+    else
+      obj = api().store_create(buf.c_str(), &c_err);
```
bnbarham, line 159: Nitpick: I'd prefer braces for the else in this case

bnbarham (commented):
Testing wise it'd be nice to add a `IndexStoreDBTests.swift` multi-language test that has some remappings + checks the paths do end up remapped.

For specifying the mappings during compilation I believe you can add to `clang_flags` and `swift_flags` (see `Documentation/Tibs.md`.

include/indexstore/IndexStoreCXX.h
```diff
@@ -42,6 +44,28 @@ class IndexStoreLibrary {
   const indexstore_functions_t &api() const { return functions; }
 };
 
+class IndexStoreCreationOptions {
```
bnbarham, line 47: IMO calling this options makes things pretty confusing. Should we just grab `PathRemapper`?

bnbarham:
> For specifying the mappings during compilation I believe you can add to clang_flags and swift_flags (see Documentation/Tibs.md.

Ah, except that you need the actual source/build paths. Which means we'd need to add something extra to tibs. Any thoughts @benlangmuir?

benlangmuir:
> > For specifying the mappings during compilation I believe you can add to clang_flags and swift_flags (see Documentation/Tibs.md.
> 
> Ah, except that you need the actual source/build paths. Which means we'd need to add something extra to tibs. Any thoughts @benlangmuir?

You could add some magic variables e.g. "$BUILD_DIR" and "$SOURCE_DIR" to tibs.  The code in `TibsBuilder.init` can do the substitution since it already needs to know the source/build directories and compute the per-target swift_flags/clang_flags.  I don't want to go overboard giving tibs macro support or anything, but this seems like a very reasonable addition.

include/indexstore/IndexStoreCXX.h
```diff
@@ -117,10 +141,22 @@ class IndexStore {
...
+  IndexStore(StringRef path, IndexStoreLibraryRef library,
+             IndexStoreCreationOptions *options,
+             std::string &error) : library(std::move(library)) {
     llvm::SmallString<64> buf = path;
     indexstore_error_t c_err = nullptr;
-    obj = api().store_create(buf.c_str(), &c_err);
+    // Backwards compatibility for previous versions which don't support a
+    // prefix mapping.
+    //
+    // TODO: Should we error if the API is unavailable but a mapping is given?
```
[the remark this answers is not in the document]
bnbarham: Just that the mapping wouldn't actually do anything, which seems odd to me.

include/CIndexStoreDB/CIndexStoreDB.h
```diff
@@ -175,6 +191,7 @@ indexstoredb_index_create(const char * _Nonnull storePath,
                   bool readonly,
                   bool enableOutOfDateFileWatching,
                   bool listenToUnitEvents,
+                  indexstoredb_creation_options_t _Nonnull options,
```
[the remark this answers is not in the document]
benlangmuir: I don't have a strong feeling about this wrapper C API -- we've never promised that this is a stable API or ABI. I would suggest either you move all the options into a single struct, or you rename the current one to just be about path remappings.  The Swift API should be stable though, so if you decide to make an options struct in Swift you should keep the existing API for compatibility.

include/indexstore/IndexStoreCXX.h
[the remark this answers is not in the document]
DavidGoldman: Yeah I think failing is probably better, so I'll add that if you think it's alright?

include/CIndexStoreDB/CIndexStoreDB.h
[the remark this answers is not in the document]
bnbarham: I'd personally like to see them all moved here. Though if stability doesn't matter we could do that in a separate PR.

I'm not sure about whether to add a new constructor for the Swift API or not. From the client side there isn't much difference, they don't need to add any argument they don't care about. Though perhaps an array of enumerators (with associated values) could make sense? Not sure if it's worth it or not 🤷

include/indexstore/IndexStoreCXX.h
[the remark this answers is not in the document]
bnbarham: I had the feeling @benlangmuir was disagreeing with me there... I assume the idea being that you'd still want to be able to use the old API without having to change the options you pass? So in that case it's better to silently ignore them than hard error.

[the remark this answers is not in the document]
benlangmuir: I'm not against an error, just making sure it make sense.  If this were a command-line tool I'd say this should be a warning.  If we wanted to be fully flexible we could have a specific error type/enum-case at the Swift level, which would allow the code that uses IndexStoreDB to decide how to handle it (e.g. emit a warning and drop the mapping option, or keep it a hard error).  On the other hand, unless someone is planning to use that flexibility maybe it's not worth it and we could just make it a normal error with a string :shrug: Errors are always an open type, so we always have the option to add more specific errors in the future without breaking source.

include/CIndexStoreDB/CIndexStoreDB.h
DavidGoldman: Done

include/indexstore/IndexStoreCXX.h
DavidGoldman: Hmm but the C API is `creation_options_add_prefix_mapping`, I think it makes sense to have this in case more options are added, right?

include/indexstore/IndexStoreCXX.h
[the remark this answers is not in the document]
DavidGoldman: I was thinking an error makes sense since if you give path remappings it presumably mean the index data uses them, thus if we don't apply them you'll get weird paths.

include/indexstore/IndexStoreCXX.h
bnbarham: I don't think the intention is that this has to match the C API. Right now the equivalent IndexStoreCXX.h in LLVM just has the PathRemapper passed. Maybe that should have had options as well, but I believe we don't care about stability there so 🤷

DavidGoldman: Ah I didn't add `PathRemapper` because that's in llvm/Basic and we don't need all of the logic that's there.

include/indexstore/IndexStoreCXX.h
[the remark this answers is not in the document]
DavidGoldman: Added an error

lib/Index/IndexSystem.cpp
```diff
@@ -641,13 +643,14 @@ IndexSystem::create(StringRef StorePath,
...
                     bool waitUntilDoneInitializing,
+                    indexstore::IndexStoreCreationOptions *options,
                     Optional<size_t> initialDBSize,
                     std::string &Error) {
   std::unique_ptr<IndexSystemImpl> Impl(new IndexSystemImpl());
   bool Err = Impl->init(StorePath, dbasePath, std::move(storeLibProvider), std::move(Delegate),
                         useExplicitOutputUnits, readonly,
                         enableOutOfDateFileWatching, listenToUnitEvents, waitUntilDoneInitializing,
-                        initialDBSize, Error);
+                        std::move(options), initialDBSize, Error);
```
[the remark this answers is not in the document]
DavidGoldman: Done

lib/CIndexStoreDB/CIndexStoreDB.cpp
```diff
@@ -94,25 +94,88 @@ class BlockIndexSystemDelegate: public IndexSystemDelegate {
...
+                          indexstoredb_creation_options_t cOptions,
                           indexstoredb_error_t *error) {
 
   auto delegate = std::make_shared<BlockIndexSystemDelegate>(delegateCallback);
   auto libProviderObj = std::make_shared<BlockIndexStoreLibraryProvider>(libProvider);
+  auto options = static_cast<CreationOptions *>(cOptions);
 
   std::string errMsg;
   if (auto index =
           IndexSystem::create(storePath, databasePath, libProviderObj, delegate,
```
DavidGoldman, line 175: @bnbarham Is this alright or should the `CreationOptions` be passed further down instead?

DavidGoldman: Done

include/indexstore/IndexStoreCXX.h
```diff
@@ -117,10 +142,26 @@ class IndexStore {
...
```
bnbarham, line 153: You've done this now :)

bnbarham (commented):
LGTM other than the couple comments I just added. I'm still not super happy with the `IndexStoreCreationOptions` and `index::CreationOptions` names but I also don't have a better suggestion. So unless @benlangmuir / @akyrtzi can think of something, it'll have to do for now.

bnbarham:
@swift-ci please test

benlangmuir (approved):
I don't have any better suggestions, and the C++ side is easy to change later if we want to

bnbarham:
@swift-ci please test

PR #151, opened 2022-05-27
Update build-script-helper.py to Python 3

ahoppen:
When running locally, it looks like we just need to change the shebang.

https://github.com/apple/indexstore-db/issues/150

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -1,6 +1,4 @@
-#!/usr/bin/env python
-
-from __future__ import print_function
+#!/usr/bin/env python3
 
 import argparse
 import os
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

Utilities/build-script-helper.py
```diff
@@ -1,6 +1,4 @@
-#!/usr/bin/env python
-
-from __future__ import print_function
+#!/usr/bin/env python3
```
bnbarham, line 1: Depending on the python 3 version I'm not sure just this change is enough. It's fine for valid args but running with no args ends up with a stacktrace instead of error - I believe this is caused by the behavior of `add_subparsers`. It should now take `required=True` in Python >= 3.7.

Utilities/build-script-helper.py
```diff
@@ -93,7 +91,7 @@ def add_common_args(parser):
     parser.add_argument('--sanitize-all', action='store_true', help='build using every available sanitizer in sub-directories of build path')
     parser.add_argument('--verbose', '-v', action='store_true', help='enable verbose output')
 
-  subparsers = parser.add_subparsers(title='subcommands', dest='action', metavar='action')
+  subparsers = parser.add_subparsers(title='subcommands', dest='action', required=True, metavar='action')
```
bnbarham, line 94: Does PR CI run on Ubuntu 18.04? 3.7 was released in 2018, so wouldn't surprise me if it didn't have it.

ahoppen: I just checked, they run Python 3.6.9 😬 I guarded the `required=True` option behind a Python version check.

bnbarham (approved):

PR #152, opened 2022-06-03
[5.7] Support for hermetic indexing + path remappings

DavidGoldman:
**Explanation**: https://github.com/apple/llvm-project/pull/4207 adds path remappings
to the index store to allow using hermetic index data but remapping
for local use without the need of any external tools like index-import.

This adds a new prefixMappings parameter to IndexStoreDB.init to allow giving the
prefix mapping to remap from hermetic/canonical --> local paths. If prefix
mappings are given while the new IndexStore API is unavailable,
indexstoredb creation will fail noting that the new API is only
available in newer versions of libIndexStore.

In addition, I've added a new `indexstoredb_creation_options_*` API to
allow configuring options when creating indexstoredb from Swift.

To support testing the remappings, TibsBuilder now supports
replacing arguments containing `$SRC_DIR` and `$BUILD_DIR` with
the respective directories.

(cherry picked from commit aff5369629596f5d1ae00676d775580a56343826)

**Scope**: Internal changes to IndexStoreDB making use of a newly added IndexStore API in the 5.7 release. IndexStoreDB constructor has changed but the new arg has a default value.

**Risk**: Low due to the test coverage below, main risk would be the new path remapping functionality.

**Testing**: Tests have been added for the remappings and pre-existing tests still pass.

### DIFF
--- a/Sources/ISDBTestSupport/INPUTS/HermeticMixedLang/a.swift
+++ b/Sources/ISDBTestSupport/INPUTS/HermeticMixedLang/a.swift
@@ -1,0 +1,9 @@
+#if os(macOS)
+func test(c: /*C:ref:swift*/C) {
+  c . /*C.method:call:swift*/method()
+}
+#endif
+
+func test() {
+  /*bridgingHeader:call*/bridgingHeader()
+}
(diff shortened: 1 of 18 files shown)


bnbarham:
@swift-ci please test

airspeedswift (approved):

PR #153, opened 2022-06-07
Add missing <ratio> header in "llvm/Support/Chrono.h"

bnbarham:
`Chrono.h` uses `ratio`, but doesn't include its corresponding header.
This is fine on main and on the regular SPM build because `chrono`
includes `ratio` in one of its private headers and main's `chrono` just
exports all the headers. But on rebranch the private headers are now ...
private and thus `ratio` is no longer visible.

Include it directly in `Chrono.h` to unblock rebranch for now.

### DIFF
--- a/include/llvm/Support/Chrono.h
+++ b/include/llvm/Support/Chrono.h
@@ -14,6 +14,7 @@
 
 #include <chrono>
 #include <ctime>
+#include <ratio>
 
 namespace llvm {
 


bnbarham:
@swift-ci please test

compnerd (approved):

benlangmuir (approved):
We should update our copy of llvm of course, but no need to block on that.

bnbarham:
> We should update our copy of llvm of course, but no need to block on that.

Yep, I'll try get all the things green and then come back to that 🙇.

PR #154, opened 2022-07-29
Several changes in build-script-helper.py to support building indexstore-db in a unified build

ahoppen:
This contains the analogous changes from https://github.com/apple/sourcekit-lsp/pull/597 as well as the following additional ones:

- Add SWIFTCI_USE_LOCAL_DEPS=1 environment variable when invoking SwiftPM
  - So far, this has not been relevant because indexstore-db didn’t have any dependencies. Once we build indexstore-db in a unified SwiftPM the `swift build` invocation will also resolve Package.swift for the other packages in the unified build, which need this environment variable.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+
+ignore =
+    E501,
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -132,10 +132,6 @@ def handle_invocation(swift_exec: str, args: argparse.Namespace) -> None:
     Depending on the action in 'args', build the package or run tests.
     """
     if args.action == 'build':
-        # Workaround for incremental build bug in swiftpm.
-        print('Cleaning ' + args.build_path)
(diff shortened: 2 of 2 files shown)


benlangmuir:
> This contains the analogous changes from [apple/sourcekit-lsp#597](https://github.com/apple/sourcekit-lsp/pull/597) as well as the following additional ones:
> 
> * Build all libraries as static libraries instead of auto
>   
>   * Otherwise, we can’t explicitly build these libraries by specifying e.g. `--product IndexStoreDB`. But we need to be able to request explicit builds once we build indexstore-db in a unified SwiftPM build.
>   * I’m open to debate whether these should be `.dynamic` instead of `.static` but static libraries appear to work at the moment.

Since we don't "install" this product anywhere currently, how about we just use the test target instead? Then we can keep the main products 'auto'.

```
--product IndexStoreDBPackageTests
```

benlangmuir (approved):

bnbarham (approved):

PR #156, opened 2022-08-19
ISDBTibs: adjust `touch` for Windows

compnerd:
Windows does not have the `touch` command.  Since we are simply creating
an empty file, we can emulate this using `copy NUL $out`.  This is
equivalent to using `cp /dev/null $out` on Unix platforms.  We can now
at least run through the index build part of the SourceKit-LSP test.


compnerd:
@swift-ci please test

bnbarham (approved):

PR #158, opened 2022-09-01
ISDBTibs: honor `SDKROOT` as default SDK

compnerd:
Windows requires a SDK to be provided.  The default SDK path is the
location specified by `SDKROOT` in the environment.  Use this value for
the default value.

### DIFF
--- a/Sources/ISDBTibs/TibsBuilder.swift
+++ b/Sources/ISDBTibs/TibsBuilder.swift
@@ -406,10 +406,12 @@ extension TibsBuilder {
 
   /// The default sdk path to use on Darwin (on other platforms, returns nil).
   public static var defaultSDKPath: String? = {
-    #if !os(macOS)
-    return nil
-    #else
+    #if os(macOS)
     return xcrunSDKPath()
+    #elseif os(Windows)
+    return ProcessInfo.processInfo.environment["SDKROOT"]
+    #else
+    return nil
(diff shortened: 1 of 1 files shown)


Sources/ISDBTibs/TibsBuilder.swift
```diff
@@ -406,10 +406,12 @@ extension TibsBuilder {
 
   /// The default sdk path to use on Darwin (on other platforms, returns nil).
```
bnbarham, line 407: Comment is now out of date 😅. Would it hurt to use the environment variable as a default for all non-macos platforms?

bnbarham (approved):

Sources/ISDBTibs/TibsBuilder.swift
compnerd: Sure, we should be able to do that as Linux will not set `SDKROOT` by default.  IIRC, unlike Windows and Darwin, Linux will effectively treat `-sdk` as `--sysroot`.

compnerd: Sure, we should be able to do that as Linux will not set `SDKROOT` by default.  IIRC, unlike Windows and Darwin, Linux will effectively treat `-sdk` as `--sysroot`.

compnerd:
@swift-ci please test

PR #160, opened 2022-09-29
Tentative fix for MDB_PAGE_FULL errors

bnbarham:
https://bugs.openldap.org/show_bug.cgi?id=9806 notes MDB_PAGE_FULL can occur for widely varying record sizes (which we definitely have). The size chosen for `nkeys` was for 4KB pages, but M1s (and up) have 16KB pages. For now set this based on the page size of the system, at least until there's an actual fix in LMDB itself.

Resolves rdar://100363010.

### DIFF
--- a/lib/Database/lmdb/mdb.c
+++ b/lib/Database/lmdb/mdb.c
@@ -9790,7 +9790,12 @@ mdb_page_split(MDB_cursor *mc, MDB_val *newkey, MDB_val *newdata, pgno_t newpgno
 			 * the split so the new page is emptier than the old page.
 			 * This yields better packing during sequential inserts.
 			 */
-			if (nkeys < 32 || nsize > pmax/16 || newindx >= nkeys) {
+// INDEXSTOREDB START
+// Changed in IndexStoreDB's copy of LMDB only. Should be removed once
+// ITS#9806 is fixed upstream. This changes the `nkeys < 32` check, which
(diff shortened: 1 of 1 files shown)


bnbarham:
@swift-ci please test

akyrtzi (approved):

lib/Database/lmdb/mdb.c
```diff
@@ -9790,7 +9790,7 @@ mdb_page_split(MDB_cursor *mc, MDB_val *newkey, MDB_val *newdata, pgno_t newpgno
 			 * the split so the new page is emptier than the old page.
 			 * This yields better packing during sequential inserts.
 			 */
-			if (nkeys < 32 || nsize > pmax/16 || newindx >= nkeys) {
+			if (nkeys < env->me_os_psize/128 || nsize > pmax/16 || newindx >= nkeys) {
```
benlangmuir, line 9793: I would suggest adding a comment here identifying that there is an IndexStoreDB-specific change and why. 

Should this be `me_psize` instead of `me_os_psize`? They're usually the same, but it's probably more correct to use the actual db page size and not the os page size.

bnbarham: Will do. And yes, it probably should be. I actually have that change locally now 😅. Was just looking whether I could add a test for this case or not.

bnbarham:
@swift-ci please test

PR #161, opened 2022-10-18
Allow relative file paths during canonicalization

DavidGoldman:
Additionally, don't apply the working dir to unit paths, instead we should treat them as verbatim identifiers for the units.

This improves support for build systems such as Bazel where the build/execution root is remapped to ., meaning all paths including unit paths become relative. Previously, this wasn't supported because the unit name internally would be remapped to the absolute path when an index prefix mapping is used, instead of preserving the canonical unit path.

Note that the latest version of libIndexStore already makes paths absolute before writing them, so the only current way to get relative file paths in the index data is with index path mappings.

cc @bnbarham

### DIFF
--- a/Tests/IndexStoreDBTests/IndexTests.swift
+++ b/Tests/IndexStoreDBTests/IndexTests.swift
@@ -227,6 +227,54 @@ final class IndexTests: XCTestCase {
   #endif
   }
 
+  /// Same as `testHermeticExplicitOutputUnits` but with remappings to make all paths relative.
+  func testHermeticRelativePathExplicitOutputUnits() throws {
+    guard let ws = try staticTibsTestWorkspace(name: "HermeticMixedLangRelativePaths", useExplicitOutputUnits: true) else { return }
+    // Provide the reverse mappings to the one in the project.json.
+    try ws.reinitIndexStore(useExplicitOutputUnits: true,
+                            prefixMappings: [
+      PathMapping(original: ".", replacement: ws.sources.rootDirectory.path),
+    ])
+    try ws.buildAndIndex()
+    let index = ws.index
+
+  #if os(macOS)
+    let cdecl = Symbol(usr: "c:objc(cs)C", name: "C", kind: .class)
+    let getOccs = { index.occurrences(ofUSR: cdecl.usr, roles: .all) }
+
+    // Output units are not set yet.
+    XCTAssertEqual(0, getOccs().count)
+  #endif
+
+    // We must use the canonical output paths, not the local ones.
+    let indexOutputPaths = ws.builder.indexOutputPaths.map {
+      $0.path.replacingOccurrences(of: ws.builder.buildRoot.path, with: ".")
+    }
+    index.addUnitOutFilePaths(indexOutputPaths, waitForProcessing: true)
+
+    // The bridging header is referenced as a PCH unit dependency, make sure we can see the data.
+    let bhdecl = Symbol(usr: "c:@F@bridgingHeader", name: "bridgingHeader", kind: .function)
+    let bridgingHeaderOccs = index.occurrences(ofUSR: bhdecl.usr, roles: .all)
+    checkOccurrences(bridgingHeaderOccs, expected: [
+      bhdecl.at(ws.testLoc("bridgingHeader:decl"), roles: .declaration),
+      bhdecl.with(name: "bridgingHeader()").at(ws.testLoc("bridgingHeader:call"), roles: .call),
+    ])
+  #if os(macOS)
+    checkOccurrences(getOccs(), expected: [
+      cdecl.at(ws.testLoc("C:decl"), roles: [.declaration, .canonical]),
+      cdecl.at(ws.testLoc("C:def"), roles: .definition),
+      cdecl.at(ws.testLoc("C:ref:swift"), roles: .reference),
+    ])
+
+    let outUnitASwift = try XCTUnwrap(indexOutputPaths.first{ $0.hasSuffix("-a.swift.o") })
+    index.removeUnitOutFilePaths([outUnitASwift], waitForProcessing: true)
+    checkOccurrences(getOccs(), expected: [
+      cdecl.at(ws.testLoc("C:decl"), roles: [.declaration, .canonical]),
+      cdecl.at(ws.testLoc("C:def"), roles: .definition),
+    ])
+  #endif
+  }
+
   func testSwiftModules() throws {
     guard let ws = try staticTibsTestWorkspace(name: "SwiftModules") else { return }
     try ws.buildAndIndex()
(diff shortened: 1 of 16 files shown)


DavidGoldman:
@swift-ci please test

lib/Support/Path.cpp
```diff
@@ -63,11 +66,16 @@ CanonicalPathCacheImpl::getCanonicalPath(StringRef Path, StringRef WorkingDir) {
       return It->second;
   }
 
+  StringRef CanonPath;
   llvm::SmallString<PATH_MAX> Buffer;
-  if (llvm::sys::fs::real_path(AbsPath.c_str(), Buffer, false)) {
-    return CanonicalFilePathRef::getAsCanonicalPath(AbsPath);
+  if (IsRelative) {
+    CanonPath = AbsPath;
```
benlangmuir, line 72: IIUC this treats all relative paths as "canonical", which seems surprising. I would expect symlinks to be resolved, redundant path components to be trimmed, etc. as with an absolute path.  Even the idea of returning a relative path from something named `getCanonicalPath` is surprising enough that I think it deserves a doc comment at least.  Implementation-wise if you wanted a canonical relative path you could form the absolute path, `real_path` it, and remove the leading working directory if it matches. If the real path is outside the working directory I would expect to get an absolute path.

bnbarham: I was expecting updates to `addUnitOutFilePaths` and `registerUnit` (possibly others) rather than changing `getCanonicalPath`. Ie. we need to stop treating the out file as a path and instead just an identifier. I'd still expect paths to be canonicalized properly though.

include/IndexStoreDB/Index/StoreUnitInfo.h
```diff
@@ -23,13 +23,13 @@ namespace index {
 struct StoreUnitInfo {
   std::string UnitName;
   CanonicalFilePath MainFilePath;
-  CanonicalFilePath OutFilePath;
+  StringRef OutFilePath;
```
bnbarham, line 26: Might be nice to rename this to something like `OutFile` or `OutFileIdentifier`, but I'm also fine with it staying the way it is.

lib/Database/ImportTransaction.cpp
```diff
@@ -100,12 +100,11 @@ IDCode ImportTransaction::Implementation::addSymbolInfo(IDCode provider, StringR
   return usrCode;
 }
 
-IDCode ImportTransaction::Implementation::addFilePath(CanonicalFilePathRef canonFilePath) {
+IDCode ImportTransaction::Implementation::addFilePath(StringRef filePath) {
```
bnbarham, line 103: Would probably be better to have two methods here and some comments explaining that the string one should only be used for the out files.

lib/Index/FileVisibilityChecker.cpp
```diff
@@ -66,10 +66,7 @@ void FileVisibilityChecker::addUnitOutFilePaths(ArrayRef<StringRef> filePaths) {
 
   ReadTransaction reader(DBase);
   for (StringRef filePath : filePaths) {
-    CanonicalFilePath canonPath = CanonPathCache->getCanonicalPath(filePath);
-    if (canonPath.empty())
-      continue;
-    IDCode pathCode = reader.getFilePathCode(canonPath);
+    IDCode pathCode = reader.getUnitPathCode(filePath);
```
bnbarham, line 69: Could also rename this depending on what the name of `OutFile` becomes. `getUnitFileCode`? That difference seems too subtle to me though.

lib/Index/FilePathIndex.cpp
```diff
@@ -165,7 +165,9 @@ bool FileIndexImpl::foreachMainUnitContainingFile(CanonicalFilePathRef filePath,
       currUnit.UnitName = unitInfo.UnitName;
       currUnit.ModTime = unitInfo.ModTime;
       currUnit.MainFilePath = reader.getFullFilePathFromCode(unitInfo.MainFileCode);
-      currUnit.OutFilePath = reader.getFullFilePathFromCode(unitInfo.OutFileCode);
+      // Roundtrip is OK here since it assumes the path is already canonicalized so it won't
+      // change the out file path at all.
+      currUnit.OutFilePath = reader.getFullFilePathFromCode(unitInfo.OutFileCode).getPath();
```
bnbarham, line 170: I'd prefer just adding another method that returns a string rather than one that returns a `CanonicalFilePathRef`, even with the comment.

lib/Support/Path.cpp
```diff
@@ -47,13 +47,16 @@ CanonicalPathCacheImpl::getCanonicalPath(StringRef Path, StringRef WorkingDir) {
     return CanonicalFilePath();
 
   SmallString<256> AbsPath;
+  bool IsRelative = false;
```
bnbarham, line 50: Just to be clear, none of the Path.cpp changes would be required after the changes to not canonicalize the out file above right?

DavidGoldman:
@swift-ci please test

DavidGoldman:
Friendly ping

lib/Database/ImportTransaction.cpp
```diff
@@ -278,7 +286,7 @@ IDCode ImportTransaction::Implementation::addUnitFileDependency(IDCode unitCode,
   auto &db = DBase->impl();
   auto &dbiUnitByFileDependency = db.getDBIUnitByFileDependency();
 
-  IDCode fileCode = addFilePath(filePathDep);
+  IDCode fileCode = addFilePath(filePathDep.getPath());
```
bnbarham, line 289: Don't need this change any more

include/IndexStoreDB/Database/ReadTransaction.h
```diff
@@ -81,6 +81,9 @@ class INDEXSTOREDB_EXPORT ReadTransaction {
   bool getFullFilePathFromCode(IDCode filePathCode, raw_ostream &OS);
   CanonicalFilePath getFullFilePathFromCode(IDCode filePathCode);
   CanonicalFilePathRef getDirectoryFromCode(IDCode dirCode);
+  /// Returns empty path if it was not found. This should only be used for the unit path since it is not treated as
+  ///  a canonicalized path.
```
bnbarham, line 85: Nitpick: extra space

Tests/IndexStoreDBTests/IndexTests.swift
```diff
@@ -227,6 +227,55 @@ final class IndexTests: XCTestCase {
   #endif
   }
 
+  /// Same as `testHermeticExplicitOutputUnits` but with a remapping to make output paths relative.
```
bnbarham, line 230: Unfortunate we have to add a whole new test just to change mappings... suppose it makes sense though.

bnbarham (approved):
LGTM other than the couple small nitpicks. Any other comments @benlangmuir?

DavidGoldman:
@swift-ci please test

bnbarham:
@DavidGoldman would you mind squashing these before I merge?

DavidGoldman:
@swift-ci please test

bnbarham:
Thanks :)

PR #163, opened 2022-11-04
Fix dangling references to the stack

bnbarham:
The recent change to out files stored references to stack-allocated strings. Change these to `std::string` instead.

Also do a bunch more renaming to `OutFileIdentifier` while I'm here.


bnbarham:
@swift-ci please test

benlangmuir (approved):

PR #165, opened 2022-12-23
build: bump the C++ standard to 14

compnerd:
This matches the changes to Package.swift.  The use of the inline initialised members for a struct in C++11 disqualify the type from being an aggregate breaking the initialiser.  The semantics were changed in C++14 which permit the construction of the `DelegateEvent`.  This repairs the build with CMake and `clang++`.


compnerd:
@swift-ci please test

bnbarham (approved):
We should probably up both to 17 to match Swift in general, but I'm fine with matching for now 👍

PR #166, opened 2023-02-07
Package.swift: bump `swift-tools-version` to 5.5

MaxDesiatov:
This allows us to use `executableTarget` for `tibs`, which resolves an issue when building IndexStoreDB with SourceKit-LSP, as the latter also needs `swift-tools-version` to be updated to 5.5. None of the CI nodes seem to be running anything older than Swift 5.5 for a long time already.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.5
 
 import PackageDescription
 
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

ahoppen (approved):

benlangmuir (approved):

PR #176, opened 2023-11-30
Avoid a copying the result for `originalFileTrigger`

hamishknight:
We can return this as a StringRef, which avoids copying when querying `OutOfDateTriggers`.

rdar://118808681


hamishknight:
@swift-ci please test

benlangmuir:
It feels mildly abstraction breaking to assume the `OutOfDateTriggerHint` can provide stable storage for the path when it's a virtual interface, so this only makes sense to me if it's a performance issue.  From what I can see, this change saves one temporary string copy per insertion into `OutOfDateTriggers`, so it presumably saves some time, but shouldn't impact overall memory use. The other user of this API will still create the same number of string copies IIUC.

PR #180, opened 2024-02-21
Reduce memory consumption for OutOfDateTriggerHints

hamishknight:
Remove `DependentUnitOutOfDateTriggerHint`, and collapse the class hierarchy into a single class `OutOfDateFileTrigger`, which stores both the file name as well as the mod time. Then, change the `OutOfDateTriggers` map in UnitMonitor to use StringRef keys. This significantly reduces memory consumption for large projects. We could likely go further here by eliminating the storing of triggers completely, but I'm leaving that as future work for now.

rdar://50929788

### DIFF
--- a/lib/Index/IndexDatastore.cpp
+++ b/lib/Index/IndexDatastore.cpp
@@ -27,6 +27,7 @@
 #include "IndexStoreDB/Support/Logging.h"
 #include "indexstore/IndexStoreCXX.h"
 #include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/StringMap.h"
 #include "llvm/ADT/StringRef.h"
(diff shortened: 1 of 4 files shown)


lib/Index/IndexDatastore.cpp
```diff
@@ -975,43 +964,63 @@ void UnitMonitor::initialize(IDCode unitCode,
...
+      // unit is out of date.
+      if (iter->second->getModTime() >= trigger->getModTime())
+        return;
+
+      // We have a newer mod time for the file, update our trigger, and inform
+      // the delegate that the unit is out of date. Note we need to overwrite
+      // the key as well since it references the path stored in the trigger.
+      *iter = {trigger->getPathRef(), trigger};
+    } else {
+      OutOfDateTriggers.insert({trigger->getPathRef(), trigger});
```
benlangmuir, line 1019: You could unconditionally call `insert` and check the `bool` in the return value to know whether it inserted or not. This avoids an unnecessary hash table lookup when the value is missing.

bnbarham: Heh, I suggested `try_emplace` but yeah, `insert` is the same here 👍

hamishknight: Went with `try_emplace` 👍

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test Windows

hamishknight:
@swift-ci please test Windows

PR #182, opened 2024-03-05
Avoid a potential use-after-move

hamishknight:
Avoid moving `unitInfo` in the body of the `for` loop here.


hamishknight:
@swift-ci please test

bnbarham (approved):

PR #190, opened 2024-04-16
Return timestamp of latest unit for file in seconds not nanoseconds

ahoppen:
The doc comment said that the timestamp was in seconds, but we were returning a timestamp in nanoseconds.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #191, opened 2024-04-16
[6.0] Return timestamp of latest unit for file in seconds not nanoseconds

ahoppen:
* **Explanation**: We were incorrectly returning a timestamp as nanoseconds instead of seconds in `indexstoredb_timestamp_of_latest_unit_for_file`
* **Scope**: The `indexstoredb_timestamp_of_latest_unit_for_file` function introduced a few weeks ago that doesn’t  have any users yet
* **Risk**: Very low, the function hasn’t been used so far
* **Testing**: Tested that this fixes an issue I was seeing in SourceKit-LSP
* **Issue**: n/a
* **Reviewer**:  @bnbarham on https://github.com/apple/indexstore-db/pull/190

### DIFF
--- a/lib/CIndexStoreDB/CIndexStoreDB.cpp
+++ b/lib/CIndexStoreDB/CIndexStoreDB.cpp
@@ -448,7 +448,7 @@ indexstoredb_symbol_location_path(indexstoredb_symbol_location_t loc) {
 double
 indexstoredb_symbol_location_timestamp(indexstoredb_symbol_location_t loc) {
   auto obj = (SymbolLocation *)loc;
-  // Up until C++20 the reference date of time_since_epoch is undefined but according to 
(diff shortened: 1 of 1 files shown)


akyrtzi:
Why not use a more precise timestamp than only up-to-second fidelity?

ahoppen:
It’s returning a `double`, so it does have sub-second fidelity.

akyrtzi:
Ah thanks for clarifying!

bnbarham (approved):
Given I had the exact same question, let's change this API to return a 64 bit int and nanoseconds instead. But we can take the fix to unblock things for now 👍

ahoppen:
Agreed to change the API to return nanoseconds in a follow-up PR.

PR #194, opened 2024-04-24
Fix issue that converted a unix timestamp to from nanoseconds to seconds twice

ahoppen:
An incorrect cherry-pick caused us to device the nanoseconds twice by 1.000.000.000, when we should only divide it once.

This makes the state of `release/6.0` match `main`.


bnbarham (approved):

ahoppen:
@swift-ci Please test

PR #198, opened 2024-06-01
Mark the closures to the `forEach` functions as non-escaping

ahoppen:
These functions don’t actually let the body escape. We should not need to mark `body` as `@escaping` and should use `withoutActuallyEscaping` in the implementation.

rdar://126803786


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #200, opened 2024-06-02
Expose the symbol provider of a `SymbolOccurrence` via the Swift API

ahoppen:


ahoppen:
@swift-ci Please test

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -289,16 +302,7 @@ public final class IndexStoreDB {
   public func symbolProvider(for sourceFilePath: String) -> SymbolProviderKind? {
```
bnbarham, line 302: IIRC this API was added for what's now using the provider from the occurrence instead, should we just remove?

bnbarham (approved):

Sources/IndexStoreDB/IndexStoreDB.swift
bnbarham: (https://github.com/apple/indexstore-db/pull/177)

ahoppen: I’ll remove it in a follow-up PR.

PR #220, opened 2024-08-12
Add size for indexstoredb_symbol_role_t

bnbarham:
This was previously miscompiling with clang (would be 32 bit instead of
64 bit) and now errors with clang 19. Note that this doesn't seem to
have had any impact within indexstore-db itself - the C++ side is
using a different enum (`SymbolRole`) which has:
```
Canonical = uint64_t(1) << 63,
```

And seems like the only real use of `Canonical` is
`foreachCanonicalSymbolOccurrenceImpl` itself (also C++).

### DIFF
--- a/include/CIndexStoreDB/CIndexStoreDB.h
+++ b/include/CIndexStoreDB/CIndexStoreDB.h
@@ -59,7 +59,7 @@ typedef void *indexstoredb_symbol_location_t;
 typedef void *indexstoredb_symbol_relation_t;
 typedef void *indexstoredb_unit_info_t;
 
-typedef enum {
+typedef enum : uint64_t {
   INDEXSTOREDB_SYMBOL_ROLE_DECLARATION = 1 << 0,
   INDEXSTOREDB_SYMBOL_ROLE_DEFINITION  = 1 << 1,
   INDEXSTOREDB_SYMBOL_ROLE_REFERENCE   = 1 << 2,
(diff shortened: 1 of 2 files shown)


bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test

include/CIndexStoreDB/CIndexStoreDB.h
```diff
@@ -59,7 +59,7 @@ typedef void *indexstoredb_symbol_location_t;
 typedef void *indexstoredb_symbol_relation_t;
 typedef void *indexstoredb_unit_info_t;
 
-typedef enum {
+typedef enum : uint64_t {
```
benlangmuir, line 62: This is only valid in C23 or later (otherwise it's a C++ extension). Clang and gcc will allow that, but I don't think MSVC does unless you're in C++ mode.

I'm not sure if there is actually a portable way to get a sized enum before C23. May need to switch to a typedef of `uint64_t`.

include/CIndexStoreDB/CIndexStoreDB.h
```diff
@@ -82,7 +82,7 @@ typedef enum {
   INDEXSTOREDB_SYMBOL_ROLE_REL_IBTYPEOF    = 1 << 17,
   INDEXSTOREDB_SYMBOL_ROLE_REL_SPECIALIZATIONOF = 1 << 18,
 
-  INDEXSTOREDB_SYMBOL_ROLE_CANONICAL = 1 << 63,
+  INDEXSTOREDB_SYMBOL_ROLE_CANONICAL = uint64_t(1) << 63,
```
benlangmuir, line 85: This is only valid in C++; you probably want `(uint64_t)1 << 63`

bnbarham: Heh, indeed. Thanks.

include/CIndexStoreDB/CIndexStoreDB.h
bnbarham: I don't think we actually use this from C anyway, only from Swift. It's theoretically possible that there's a client that is using it directly, but let's try this for now and then convert to typedef + enum + macro for the > 32 bit case if we end up having to.

bnbarham:
@swift-ci please test

bnbarham:
@swift-ci please test Windows platform

bnbarham:
@swift-ci please test Windows platform

ahoppen (approved):

PR #222, opened 2024-10-15
Fix issue that caused database rename to and from `saved` to fail

ahoppen:
`llvm::sys::fs::rename` cannot rename directories becasue it creates a file handle for the file to rename, which doesn't work for directories. Implement a custom rename function that uses `MoveFileW` on Windows. 

Fixes https://github.com/swiftlang/sourcekit-lsp/issues/1750

### DIFF
--- a/lib/Database/Database.cpp
+++ b/lib/Database/Database.cpp
@@ -19,13 +19,16 @@
 #include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/StringMap.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/Support/Errc.h"
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Mutex.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/raw_ostream.h"
+#include "llvm/Support/WindowsError.h"
 #if defined(_WIN32)
 #define NOMINMAX
 #include "Windows.h"
+#include "winbase.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
https://github.com/swiftlang/swift/pull/77015

@swift-ci Please test Windows

bnbarham (approved):
Pretty weird but... 🤷

ahoppen:
For the record, `llvm::sys::fs::rename` tries to create a file handle for the directory using `CreateFileW` and that fails. From [Windows documentation](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew) of `CreateFileW`:

> Creates or opens a file or I/O device. The most commonly used I/O devices are as follows: file, file stream, directory,

> To open a directory using CreateFile, specify the FILE_FLAG_BACKUP_SEMANTICS flag as part of dwFlagsAndAttributes

> FILE_FLAG_BACKUP_SEMANTICS: The file is being opened or created for a backup or restore operation. The system ensures that the calling process overrides file security checks when the process has SE_BACKUP_NAME and SE_RESTORE_NAME privileges. For more information, see [Changing Privileges in a Token](https://learn.microsoft.com/en-us/windows/desktop/SecBP/changing-privileges-in-a-token).

So, looks like `CreateFileW` can open directories but it’s not really supposed to except for backup purposes and should not be the go-to solution for renaming directories.

PR #223, opened 2024-10-25
On Windows use `RemoveDirectoryW` to delete empty directory before renaming directory to it

ahoppen:
I don't know why but using `llvm::sys::fs::remove_directories` to delete the directories sometimes does not actually delete the directory, which causes `MoveFileW` to fail because the destination already exists. I don't see this issue when using `RemoveDirectoryW`


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #227, opened 2024-11-11
Disable UBSAN on macOS

ahoppen:
Building indexstore-db with ubsan enabled with an Xcode 15.1 SDK causes C++ exception to not be catchable with `try {} catch {}` blocks and indexstore-db relies on exception catching, possibly related to  https://stackoverflow.com/questions/51673343/clang-with-libc-exceptions.

Disable UBSAN for now and re-enable it when CI nodes has been updated to Xcode 16.


ahoppen:
@swift-ci Please test

bnbarham (approved):

PR #236, opened 2025-02-21
Avoid crash when querying in non-visible files

matthewseaman:
Check if we weren't able to find a visible provider for the file and if not, continue checking other provider codes.


bnbarham:
@swift-ci please test

bnbarham (approved):
Thanks!

PR #247, opened 2025-05-23
Use full module triple in swiftmodule install

etcwilde:
This patch updates the CMake build to install the IndexStoreDB module with the full triple. This allows us to query the compiler for the appropriate module name instead of maintaining the long list of host architectures, making the build more resilient to changes in the future.


compnerd (approved):

bnbarham:
Could you run a Windows toolchain build (might have to cross-PR from Swift) 🙏?

PR #248, opened 2025-05-27
[indexstore-db] Declare Codable conformance, to autosynthesize implementation.

moshe-foreflight:
This PR adds `Codable` support to the `SymbolOccurrence` and `Symbol` models, in addition to types they depend on. (`SymbolRole`, etc) 

**Why?** 
- I found a need for this while trying to work with actors to parallelize index access.
- Clients code depending on `indexstore-db` would need to manually implement conformance, because conformances in extensions are not synthesized.

**Risks** 
- Additions to these models would need to support `Codable` as well. (Compiler will catch this.)
- Changes may cause unexpected results when deserializing. (Mitigated with unit tests)

### DIFF
--- a/Sources/IndexStoreDB/IndexStoreDB.swift
+++ b/Sources/IndexStoreDB/IndexStoreDB.swift
@@ -38,7 +38,7 @@ public struct PathMapping: Equatable {
   }
 }
 
-public enum SymbolProviderKind: Sendable {
+public enum SymbolProviderKind: Sendable, Codable {
   case clang
   case swift
 
--- a/Sources/IndexStoreDB/IndexStoreDBError.swift
+++ b/Sources/IndexStoreDB/IndexStoreDBError.swift
@@ -15,7 +15,7 @@ import IndexStoreDB_CIndexStoreDB
 
 import Foundation
 
-public enum IndexStoreDBError: Error {
+public enum IndexStoreDBError: Error, Codable {
   case create(String)
   case loadIndexStore(String)
 }
--- a/Sources/IndexStoreDB/Symbol.swift
+++ b/Sources/IndexStoreDB/Symbol.swift
@@ -13,7 +13,7 @@
 @_implementationOnly
 import IndexStoreDB_CIndexStoreDB
 
-public enum IndexSymbolKind: Hashable, Sendable {
+public enum IndexSymbolKind: Hashable, Sendable, Codable {
   case unknown
   case module
   case namespace
@@ -45,14 +45,14 @@ public enum IndexSymbolKind: Hashable, Sendable {
   case commentTag
 }
 
-public enum Language: Hashable, Sendable {
+public enum Language: Hashable, Sendable, Codable {
   case c
   case cxx
   case objc
   case swift
 }
 
-public struct Symbol: Hashable, Sendable {
+public struct Symbol: Hashable, Sendable, Codable {
 
   public var usr: String
   public var name: String
--- a/Sources/IndexStoreDB/SymbolLocation.swift
+++ b/Sources/IndexStoreDB/SymbolLocation.swift
@@ -14,7 +14,7 @@
 import IndexStoreDB_CIndexStoreDB
 import Foundation
 
-public struct SymbolLocation: Equatable, Sendable {
+public struct SymbolLocation: Equatable, Sendable, Codable {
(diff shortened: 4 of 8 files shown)


moshe-foreflight:
I've added some tests to validate the input and output but an argument could be made that the tests merely confirm the behavior of the system framework. Looking for input from reviewers.

ahoppen:
Thanks for opening the PR, @moshe-foreflight. 

I don’t quite understand why we need a `Codable` conformance for these types. You said that you need it to work with these types across actor instances but wouldn’t that mean that `Sendable` is the protocol you are looking for? I’m not sure what you’re doing on your end but with the `Codable` conformance you are likely serializing the data to JSON just to deserialize it in the same process, which seems like a waste of compute resources. If `Codable` is indeed the protocol you need, could you elaborate a little more what your use case is?

ahoppen:
> I'm trying to optimize for performance across multiple runs of a process by splitting the IndexStoreDB into a separate XPC process.

What are you trying to gain by keeping the index alive? Are you expecting to quit and re-launch clients of your index service multiple times per second or something like that? Because when closing the index, it should be saved (look for a `saved` directory in the directory that contains your indexstore-db) and re-opening the index again should just be re-loading that saved index, which should be almost instant.

> Quite possibly - it's likely I'm "holding it wrong."

I didn’t realize that you were talking about separate processes. In that case, `Sendable` won’t be sufficient.

> Can you think of a reason _not_ to support this from the perspective of the library vendor? Just because we can, doesn't mean we should, but will adding Codable here worsen IndexStoreDB?

I am a still a little hesitant, in particular until I understand that this is the correct solution for a real problem. Here are a couple reasons why:
- If we add the ability to dump the symbols as JSON, there might be expectations that the JSON dump stays backwards-compatible and this could significantly hinder the evolution of indexstore-db.
- The index can be quite large and dumping all of it as JSON can quickly produce dumps that are too large to handle. JSON might not be the correct dump format anymore. At that point, you would likely be better of just iterating over the entire index store itself, without using indexstore-db (ie. `dlopen` `libIndexStore.dylib` and then use these functions to iterate over the index https://github.com/swiftlang/indexstore-db/blob/main/Sources/IndexStoreDB_Index/include/IndexStoreDB_Index/indexstore_functions.h)
- Adding `Codable` conformances will prevent us from adding any kind of circular references to these types. To be clear, we don’t have any plans to do this, but I think it’s not completely unreasonable that `Symbol` became a reference type and `SymbolOccurrence` contained a pointer to `Symbol` and then `Symbol` contained a member that referred to all of its references. This wouldn’t be representable in JSON.
- In many cases, I would expect that not all the information of these types is actually needed, in which case serializing all of it would add unnecessary overhead. In these cases, it might be more beneficial to do as much processing as possible inside the process that has loaded the index and then just return the aggregated data. FWIW, this is effectively what SourceKit-LSP does to serve its LSP requests.

ahoppen:
> * At least while developing, I want to be able to iterate quickly. (XPC requests against a living service seems faster because it avoids loading even a saved index.)

Out of curiosity: How long does it take to load the saved index? I would expect it to take significantly less than a second, especially if you use a small-ish project for local development testing.

> * When my application crashes, the `saved` directory doesn't appear to be created. By keeping the index outside the main process, we don't need it.

I think I remember there being a bug where we don’t recover the indexstore-db after a crash. If I remember correctly, what we’re supposed to do is, if there is no `saved` directory, check if we have one that matches a pid, check if that process is still alive, and if not, page that indexstore-db in. We should just fix this issue. If you would be interested to look at it, that would be amazing!

bnbarham:
> I think I remember there being a bug where we don’t recover the indexstore-db after a crash

This is unfortunately intentional, as it's possible it could be in a corrupted state after a crash. I'd expect creating the database for a small indexstore to be fairly fast regardless though.

moshe-foreflight:
@ahoppen I've opened #251 to add sendability. (Also addresses #250.)

ahoppen:
> If you have a code-pointer, if Ben agrees (see his followup below) and my manager approves, I'm happy to look.

Sorry, I missed something as @bnbarham pointed out. Let’s not change it then. 

And thanks for opening the PR.

moshe-foreflight:
Closing this one, then. Thanks for the patient back-and-forth!

PR #270, opened 2026-01-09
Remove files used for test discovery on Linux

ahoppen:
We used to need these files when Linux did not have automatic test discovery. That has changed quite a while ago and these files should no longer be needed.


ahoppen:
@swift-ci Please test

ahoppen:
Oh, seems like we weren’t executing `IndexStoreTests` so far in CI. Fixing their test failure in https://github.com/swiftlang/indexstore-db/pull/277.

hamishknight (approved):

ahoppen:
@swift-ci Please test Linux

PR #285, opened 2026-04-05
Create an executable to dump index or record files

ahoppen:
This has a similar function as `c-index-test core -print-unit` and `c-index-test core -print-record` but doesn’t require building all of LLVM to use.

Fixes #264

### DIFF
--- a/Sources/ISDBTibs/Process.swift
+++ b/Sources/ISDBTibs/Process.swift
@@ -22,30 +22,23 @@ extension Process {
   /// Runs a subprocess and returns its output as a String if it has a zero exit.
   package static func tibs_checkNonZeroExit(
     arguments: [String],
-    environment: [String: String]? = nil
+    environment: [String: String]? = nil,
+    workingDirectory: URL? = nil
   ) throws -> String {
     let p = Process()
     let out = Pipe()
     let err = Pipe()
 
-    if #available(macOS 10.13, *) {
-      p.executableURL = URL(fileURLWithPath: arguments[0], isDirectory: false)
-    } else {
-      p.launchPath = arguments[0]
-    }
-
+    p.executableURL = URL(fileURLWithPath: arguments[0], isDirectory: false)
     p.arguments = Array(arguments[1...])
     if let environment = environment {
       p.environment = environment
     }
     p.standardOutput = out
     p.standardError = err
+    p.currentDirectoryURL = workingDirectory
 
-    if #available(macOS 10.13, *) {
-      try p.run()
-    } else {
-      p.launch()
-    }
+    try p.run()
 
     let dataOut = out.fileHandleForReading.readDataToEndOfFile()
     let dataErr = err.fileHandleForReading.readDataToEndOfFile()
--- a/Sources/index-dump/index-dump.swift
+++ b/Sources/index-dump/index-dump.swift
@@ -1,0 +1,112 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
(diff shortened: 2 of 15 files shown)


ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/index-dump/index-dump.swift
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+      while !indexStorePath.isRoot {
+        if FileManager.default.fileExists(atPath: indexStorePath.appending(component: "v5").filePath) {
+          return indexStorePath
+        }
+        indexStorePath.deleteLastPathComponent()
+      }
+
+      throw ValidationError("Could not infer store path. Please specify --index-store.")
```
rintaro: Worth mentioning this expects a directory containing `v5`?

Sources/index-dump/index-dump.swift
```diff
@@ -0,0 +1,120 @@
...
+    } else if let record = try? store.record(named: unitOrRecordName) {
+      print(record)
+    } else {
+      throw ValidationError("Could not find unit or record named \(unitOrRecordName) in \(indexStorePath)")
+    }
+  }
+
+  // MARK: - Mode Inference
+
+  private func inferMode(from path: String) throws -> Mode {
```
rintaro: Unused?

Sources/ISDBTibs/Process.swift
```diff
@@ -40,6 +41,7 @@ extension Process {
     }
     p.standardOutput = out
     p.standardError = err
+    p.currentDirectoryURL = workingDirectory
 
     if #available(macOS 10.13, *) {
       try p.run()
```
rintaro: While you are here, can we remove these `#available(macOS 10.13, *)` now? Otherwise `currentDirectoryURL` is `10.13` or later, so it should be:
```
if #available(macOS 10.13, *) {
  p.currentDirectoryURL = workingDirectory
} else if let workingDirectory  {
  p.currentDirectoryPath = workingDirectory.path
}
```

rintaro: Ah this package is already `platforms: [.macOS(.v14)],`. We should just remove the `10.13` checks.

Sources/index-dump/index-dump.swift
ahoppen: Good call. Changed the message to 

> Could not infer store path by a parent directory named v5. Please specify --index-store.

Sources/index-dump/index-dump.swift
ahoppen: Yes, leftover from the PR I took over. Thanks for spotting.

Sources/ISDBTibs/Process.swift
ahoppen: Well spotted. Cleaned it up.

ahoppen:
@swift-ci Please test

ahoppen:
@swift-ci Please test Windows

Sources/index-dump/index-dump.swift
```diff
@@ -0,0 +1,111 @@
...
+
+  enum Mode: String, ExpressibleByArgument {
+    case unit, record
+  }
+
+  var indexStorePath: URL {
+    get throws {
+      var indexStorePath =
+        if let explicit = indexStore {
+          URL(fileURLWithPath: explicit)
```
rintaro: Just to confirm. It is intentional not to return this immediately, right?
If so I feel "Please specify --index-store" sounds strange if the path manually specified by `--index-store` doesn't contain `v5`.

rintaro (approved):

Sources/index-dump/index-dump.swift
ahoppen: No, returning directly would be a better option in case your index store path doesn’t have a v5 directory for some reason. Also adjusted the error message again a little bit.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

PR #286, opened 2026-04-15
Fix non-deterministic test failures on Amazon Linux 2 due to incorrectly set working directories

ahoppen:
On Amazon Linux 2 `_CFPosixSpawnFileActionsChdir` is not implemented because it AL2 doesn't support posix_spawn_file_actions_addchdir_np. Because of this, swift-corelibs-foundation changes the working directory of the current process.

https://github.com/swiftlang/swift-corelibs-foundation/blob/0b23e798e921fd14ff59b37c9145021cd60e0aa9/Sources/Foundation/Process.swift#L989

If we run multiple sub-processes concurrently, this races to change global state, resulting in non-deterministic working directories and spurious test failures. Instead set the working directory by passing `-working-directory` to the indexing processes.

rdar://174708032

### DIFF
--- a/Sources/ISDBTibs/Process.swift
+++ b/Sources/ISDBTibs/Process.swift
@@ -22,8 +22,7 @@ extension Process {
   /// Runs a subprocess and returns its output as a String if it has a zero exit.
   package static func tibs_checkNonZeroExit(
     arguments: [String],
-    environment: [String: String]? = nil,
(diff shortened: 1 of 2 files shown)


ahoppen:
@swift-ci Please test

hamishknight (approved):

PR #287, opened 2026-04-24
Support '.swiftinterface' in forEachMainFileContainingFile

rintaro:
When resolving main files for a `.swiftinterface` or `.swiftmodule` path, the cross-language filter previously treated these paths as non-Swift (only `.swift` was recognised), causing all Swift main files to be skipped. Extend the `isSwift` check to include `.swiftinterface` and `.swiftmodule` so that callers such as `mainFiles(containing:)` can correctly resolve source files that import a given module.

### DIFF
--- a/Sources/IndexStoreDB/IndexStoreDB.swift
+++ b/Sources/IndexStoreDB/IndexStoreDB.swift
@@ -303,12 +303,17 @@ public final class IndexStoreDB {
     crossLanguage: Bool,
     body: (String) -> Bool
   ) -> Bool {
+    let isSwift = if crossLanguage {
+      // If 'crossLanguage' is requested, this value is not used.
+      false
+    } else {
+      path.hasSuffix(".swift") || path.hasSuffix(".swiftinterface") || path.hasSuffix(".swiftmodule")
+    }
     let fromSwift = path.hasSuffix(".swift")
     return withoutActuallyEscaping(body) { body in
       return indexstoredb_index_units_containing_file(impl, path) { unit in
         let mainFileStr = String(cString: indexstoredb_unit_info_main_file_path(unit))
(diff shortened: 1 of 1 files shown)


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test macOS

Sources/IndexStoreDB/IndexStoreDB.swift
```diff
@@ -303,12 +303,17 @@ public final class IndexStoreDB {
     crossLanguage: Bool,
     body: (String) -> Bool
   ) -> Bool {
+    let isSwift = if crossLanguage {
+      // If 'crossLanguage' is requested, this value is not used.
+      false
+    } else {
+      path.hasSuffix(".swift") || path.hasSuffix(".swiftinterface") || path.hasSuffix(".swiftmodule")
+    }
```
ahoppen, line 311: The check seems fast enough that we don’t need the extra complexity of skipping the computation when `crossLanguage == false`

ahoppen (approved):

hamishknight (approved):

AnthonyLatsis:
> When resolving main files for a `.swiftinterface` or `.swiftmodule` path

What’s a main file here?

rintaro:
Main file here means a compilation unit that imports/includes the specified file.
E.g. `Foo.cpp` file for `Foo.h` file, `Some.swift` for `Swift.swiftmodule/<trilple>.swiftinterface` file, etc.
This has been used for inferring the build settings for `.h` files.

AnthonyLatsis:
Thanks, Rintaro. I really think we should call it `forEachCompilationUnitImportingFile` or similar for the sake of clarity.

PR #291, opened 2026-07-07
Re-import explicitly processed units even if their mtime is unchanged

rintaro:
`processUnitsForOutputPathsAndWait` is the explicit "I just regenerated these units, load them now" API that clients like SourceKit-LSP call after re-indexing. Its effect is silently dropped when the rewritten unit's modification time exactly equals the one in the database: `UnitDataImport` treats the unit as up to date and never re-reads it. This is easy to hit when a unit is rewritten within one file-system timestamp tick, especially on coarse-granularity file systems — where we've seen the resulting flakiness on Windows (a workspace symbol query returning 0 results right after indexing reports finished).

This threads an `ignoreModTime` flag from `processUnitsForOutputPathsAndWait` down to `UnitDataImport` so units processed through this explicit API are always re-imported. The flag is applied before the up-to-date short-circuit, so the previous-dependency bookkeeping used for cleanup is still populated. All paths that only *guess* whether a unit changed (file watching, `pollForUnitChangesAndWait`, the initial scan) keep comparing modification times, so this adds no extra work when scanning.

Speculative fix for rdar://178781218 / swiftlang/sourcekit-lsp#2687. I have no captured failing log, so the same-tick collision is a strong inference; the definitive check is to point sourcekit-lsp's `indexstore-db` dependency at this branch and stress-loop `testEnsureSymbolsLoadedIntoIndexstoreDbWhenIndexingHasFinished`, ideally on Windows.

### DIFF
--- a/Sources/IndexStoreDB_Database/ImportTransaction.cpp
+++ b/Sources/IndexStoreDB_Database/ImportTransaction.cpp
@@ -402,7 +402,8 @@ void ImportTransaction::commit() {
   return Impl->commit();
 }
 
-UnitDataImport::UnitDataImport(ImportTransaction &import, StringRef unitName, llvm::sys::TimePoint<> modTime)
+UnitDataImport::UnitDataImport(ImportTransaction &import, StringRef unitName, llvm::sys::TimePoint<> modTime,
+                               bool ignoreModTime)
 : Import(import), UnitName(unitName), ModTime(modTime), IsSystem(false) {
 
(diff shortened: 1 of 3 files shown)


rintaro:
@swift-ci Please test

ahoppen (approved):
We had a similar issue like the one you described before (https://github.com/swiftlang/sourcekit-lsp/pull/2091), so the analysis might be correct. Unfortunate that we need to do this but such is life.

hamishknight (approved):

PR #293, opened 2026-08-10
Fix foreachSymbolInFilePath/foreachSymbolOccurrenceInFilePath to visit all units of a shared file

### LINKED ISSUE #292: foreachSymbolOccurrenceInFilePath (and symbolOccurrences(inFilePath:)) returns occurrences from only one compiling target when a file belongs to multiple units
## Summary

`IndexStoreDB.symbolOccurrences(inFilePath:)` (and `symbols(inFilePath:)`, which shares the same
underlying enumeration) silently returns occurrences from only **one** of the compilation units
that reference a given source file, when that file is a member of more than one unit (e.g. an app
target and an extension target both compiling the same shared `.swift` file — a routine iOS/macOS
architecture: app + notification-service extension, app + widget extension, app + share extension,
or any multi-target setup with shared sources).

## Environment

- `indexstore-db` revision: `003ac41513ba291f10ff1a0147ae68588914668d` (pinned, matches
  `swift/release/6.3`'s current HEAD at time of writing)
- Toolchain: Xcode 26.4 / Swift 6.3, macOS (arm64)
- Found while building [`swift-isolation-map`](https://github.com/btctcn/swift-isolation-map), a
  static actor-isolation analysis CLI that reads `IndexStoreDB` for its call graph, against a real,
  ~2251-file production iOS app with two notification-extension targets sharing source files with
  the main app target.

## Minimal reproduction

No SwiftPM, no Xcode project, no third-party dependency -- two `swiftc` invocations against the
same file, indexed into one store:

```sh
SDK=$(xcrun --sdk macosx --show-sdk-path)
TARGET="arm64-apple-macosx13.0"

cat > Shared.swift <<'EOF'
public struct SharedThing {
    public init() {}
    public func doWork() { helper() }
    private func helper() { print("working") }
}
EOF

xcrun swiftc -module-name AppModule -swift-version 6 -sdk "$SDK" -target "$TARGET" \
  -emit-module -emit-module-path AppModule.swiftmodule -parse-as-library \
  -index-store-path "$PWD/index-store" -c Shared.swift -o AppModule.o

xcrun swiftc -module-name ExtModule -swift-version 6 -sdk "$SDK" -target "$TARGET" \
  -emit-module -emit-module-path ExtModule.swiftmodule -parse-as-library \
  -index-store-path "$PWD/index-store" -c Shared.swift -o ExtModule.o
```

Then, opening `index-store` with `IndexStoreDB` and calling
`symbolOccurrences(inFilePath: "<absolute path to Shared.swift>")`:

- **Expected**: 8 definitions (4 from `AppModule`'s compilation + 4 from `ExtModule`'s).
- **Actual**: 4 definitions -- only one module's worth (which one is not deterministic from the
  caller's perspective; depends on `foreachUnitContainingFile`'s own internal enumeration order).

Cross-checked directly against `libIndexStore`'s own raw C API
(`indexstore_record_reader_occurrences_apply_f` on each of the two distinct on-disk records for
this file) -- the raw index data is complete and correct; both modules' occurrences are present on
disk. The loss happens specifically in `IndexStoreDB`'s own file-based enumeration, not in the
underlying index store.

## Root cause

`Sources/IndexStoreDB_Index/SymbolIndex.cpp`,
`SymbolIndexImpl::foreachSymbolOccurrenceInFilePath` (lines 509-534):

```cpp
bool SymbolIndexImpl::foreachSymbolOccurrenceInFilePath(CanonicalFilePathRef filePath,
                                                        function_ref<bool(SymbolOccurrenceRef Occur)> Receiver) {
  bool didFinish = true;
  ReadTransaction reader(DBase);

  IDCode filePathCode = reader.getFilePathCode(filePath);
  reader.foreachUnitContainingFile(filePathCode, [&](ArrayRef<IDCode> idCodes) -> bool {
    for (IDCode idCode : idCodes) {
      UnitInfo unitInfo = reader.getUnitInfo(idCode);
      for (UnitInfo::Provider provider : unitInfo.ProviderDepends) {
        IDCode providerCode = provider.ProviderCode;
        if (provider.FileCode == filePathCode) {
          auto record = createVisibleProviderForCode(providerCode, reader);
          if (!record) {
            continue;
          }
          didFinish = record->foreachSymbolOccurrence(Receiver);

          return false;   // <-- returns from the OUTER foreachUnitContainingFile callback
        }
      }
    }
    return true;
  });
  ...
}
```

The `return false;` on the line marked above returns from the *outer* callback passed to
`reader.foreachUnitContainingFile`, not just the inner `for` loops -- so as soon as the **first**
unit containing this file yields a matching provider, the entire unit enumeration stops. Any other
unit that also contains this file (i.e. every other target compiling it) is never visited, and its
occurrences are silently never reported to `Receiver`.

The same shape likely affects `foreachSymbol`/`symbols(inFilePath:)` too, since it presumably
shares comparable enumeration logic -- not independently verified in this report, called out here
so it can be checked in the same pass as the fix.

## Impact

Any tool built on `symbolOccurrences(inFilePath:)`/`symbols(inFilePath:)` -- code navigation,
find-references, refactoring tooling, static analysis -- silently loses data for any source file
shared across more than one compilation target, with no error, warning, or partial-result
indication. This is not a rare shape: app + notification/widget/share extension sharing source
files is routine iOS/macOS architecture. Measured on one real, ~2251-file production app: a single
shared file compiled into 3 targets (main app + 2 notification extensions) had only 1 of 3 targets'
worth of declarations (72 of 216, 33%) visible via `symbolOccurrences(inFilePath:)`; the same
shape's call-graph queries lost 209 of 627 call sites (33%) for the same reason.

## Suggested fix

Continue the outer `foreachUnitContainingFile` enumeration across all matching units/providers
instead of returning `false` after the first match -- accumulating occurrences from every
compiling target's own record for this file path, not just the first one found. Happy to submit a
PR with a fix and a regression test if that's useful.

## Independently re-confirmed

Re-ran the exact reproduction above on a separate build (fresh working directory, same toolchain), calling `IndexStoreDB.symbolOccurrences(inFilePath:)` directly against the resulting store: **4** definitions returned (`AppModule` only), confirming this is deterministic and not specific to the first environment it was found in.

btctcn:
## Summary

Fixes #292.

`symbolOccurrences(inFilePath:)` and `symbols(inFilePath:)` silently return occurrences from only **one** of the compilation units that reference a given source file, when that file is a member of more than one unit (e.g. an app target and an extension target both compiling the same shared `.swift` file, or any header shared by multiple translation units).

### Root cause

In `SymbolIndexImpl::foreachSymbolInFilePath` and `SymbolIndexImpl::foreachSymbolOccurrenceInFilePath` (`Sources/IndexStoreDB_Index/SymbolIndex.cpp`), the callback passed to `reader.foreachUnitContainingFile` returned `false` right after processing the *first* matching provider it found. That `false` terminates the *entire* unit enumeration, not just the inner per-unit provider loop — so any other unit that also contains the file is never visited, and its occurrences are silently dropped.

### Fix

Keep enumerating the remaining units for the file instead of stopping after the first match, while still breaking out of the inner per-unit provider loop once a visible provider for that unit is found (so a single unit doesn't contribute duplicate occurrences), and still stopping the whole search early if the receiver itself requests it.

### Tests

- Added a new `SharedFileMultiTarget` tibs fixture: two independent targets (`TargetA`, `TargetB`) each independently compile the same `shared.swift`, mirroring the real-world app + extension-target shape from the issue.
- Added `testSymbolsInFileSharedAcrossMultipleTargets` / `testSymbolOccurrencesInFileSharedAcrossMultipleTargets`, which fail on unpatched `main` (returning 4 instead of the expected 8 occurrences) and pass with the fix.
- Updated the existing `testSymbolsInFileC` expectation: `d.h` is `#include`d by both `d.c` and `e.c` in the `CProject` fixture, so its declaration legitimately belongs to two translation units and is now correctly reported twice instead of being silently deduplicated to one.

Verified full `swift test` run (`IndexStoreDBTests` + `ISDBTibsTests` + `IndexStoreTests`) green on `main` with a matching toolchain (`swift-DEVELOPMENT-SNAPSHOT-2026-07-11-a`), with no reduction from the pre-fix baseline.

## Test plan

- [x] `swift build` succeeds
- [x] `swift test` passes in full (62/62), including the two new regression tests
- [x] Confirmed the two new tests fail on unpatched `main` and pass with the fix
- [x] Confirmed no other existing test regressed


ahoppen (commented):
High-level question: What’s the motivation for returning the occurrences from both units here? I suppose both occurrences are at the exact same location and they share the same properties, so reporting it only once for eg. find-references should be sufficient. Or am I missing a use case?

btctcn:
Good question — here's a concrete way to see what's actually happening, rather than reasoning about it abstractly.

I built the minimal repro from the issue (`Shared.swift` compiled into `AppModule` and `ExtModule`, one shared index store) and queried the same on-disk store two ways: once through `libIndexStore`'s own raw C API directly, and once through `IndexStoreDB.symbolOccurrences(inFilePath:)` on `main` (unpatched):

```
raw libIndexStore:        8 occurrences (4 from AppModule, 4 from ExtModule)
IndexStoreDB (unpatched):  4 occurrences (ExtModule only — AppModule's copy is gone entirely)
```

The raw index data on disk is complete and correct for both targets — this isn't a case of two occurrences racing to report the "same" fact twice. `IndexStoreDB`'s own file-based enumeration is silently discarding one whole target's data before it ever reaches the caller. Since this repo is the API layer between `libIndexStore` and every consumer, the question isn't really "should call sites be reported once or twice" — it's "should `symbolOccurrences(inFilePath:)` silently return a strict subset of what's actually on disk with no signal that anything was dropped." Right now it does, non-deterministically (which unit "wins" depends on `foreachUnitContainingFile`'s internal enumeration order, not anything the caller controls).

On the real-world motivating case (app + notification-extension targets sharing source), this wasn't a cosmetic gap either — it meant 33% of call sites (209 of 627) referencing symbols in the shared file were invisible to any caller working from `symbolOccurrences(inFilePath:)`/`symbols(inFilePath:)`.

PR #294, opened 2026-08-11
Clear the unit name buffer in processUnitsForOutputPathsAndWait

rintaro:
`getUnitNameFromOutputPath` appends to the name buffer, so without clearing it between iterations only the first output path of a batch produced a valid unit name; the concatenated names of the other entries match no unit and were silently dropped. Present since 2651f64. sourcekit-lsp calls this with a whole indexing batch (up to 25 Swift files), so all but the first file only became queryable once file watching caught up.

Two tests index more than one file in a single batch, one on a fresh build and one re-processing after edits; both fail without the fix. They look units up by resolved output paths, since the compiler resolves symlinks when making a relative output path absolute, so `URLUtils.swift` moves from `IndexStoreTests` to `ISDBTibs` to share `URL.realpath`. The public entry points now document that requirement.


rintaro:
@swift-ci Please test

Sources/IndexStoreDB_Index/IndexDatastore.cpp
```diff
@@ -757,6 +757,7 @@ void StoreUnitRepo::processUnitsForOutputPathsAndWait(ArrayRef<StringRef> output
   SmallString<128> nameBuf;
   std::vector<UnitEventInfo> events;
   for (StringRef outputPath : outputPaths) {
+    nameBuf.clear();
```
hamishknight: IMO it would be cleaner to just move the `nameBuf` declaration into the loop body

hamishknight (approved):
Good catch!

ahoppen (approved):
Nice catch. Thanks for fixing.

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

rintaro:
@swift-ci Please test Windows

